PR #13, opened 2024-10-10
Tech322/habit

techeng322:


app/api/chat/route.ts
```diff
@@ -5,6 +5,8 @@ import {
 import { enhanceRouteHandler } from "@/packages/next/src/routes";
 
 export const dynamic = "force-dynamic";
+export const revalidate = 0;
+export const fetchCache = "force-no-store";
```
SweetmanTech, line 9: 💯 🧠

lib/chat/getChatContext.tsx
```diff
@@ -13,64 +13,72 @@ import getRecentScore from "./getRecentScore";
 import getScoresInPast24 from "./getScoresInPast24";
 import getFans from "./getFans";
 
-const getChatContext = async () => {
+const getChatContext = async (isHabitQuestion: boolean) => {
```
SweetmanTech, line 16: I like how you're thinking of creating different `instructions` based on which `function` the user is intending to `invoke`.

Would it be a more scalable approach to use `function calling`?
- actual: `habits` is only triggered on `suggestion pill click`
- expected: AI is used to determine if the `habitQuestion` is being invoked.

https://openai.com/index/function-calling-and-other-api-updates/

Screenshot 2024-10-10 at 4 36 05 PM

PR #15, opened 2024-10-10
Tech322/notes

techeng322:


lib/server/chat-llm.service.ts
```diff
@@ -68,7 +68,7 @@ class ChatLLMService {
     // we need to limit the history length so not to exceed the max tokens of the model
     // let's assume for simplicity that all models have a max tokens of 128000
     // so we need to make sure that the history doesn't exceed output length + system message length
-    const maxModelTokens = 128000;
+    const maxModelTokens = 228000;
```
SweetmanTech, line 71: Are you confident the current model we are using supports `228000` tokens???

SweetmanTech (approved):

PR #42, opened 2024-10-29
Tech322/create new artist

techeng322:


components/Tools/Content.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface ContentProps {
+  toolName: string | undefined;
+  context: Record<string, unknown>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  fans: Array<any>;
+  scroll: () => void;
+}
+
+const Content = ({ toolName, context, fans, scroll }: ContentProps) => (
```
SweetmanTech: Could you please name this component `<ToolContent />` to be more explicit that this is used for `function tools`?
```suggestion
const ToolContent = ({ toolName, context, fans, scroll }: ContentProps) => (
```

lib/tools/createArtist.tsx
SweetmanTech, line 1: Thank your for creating modular libs for each tool call. 🙏

lib/tools/createArtist.tsx
```diff
@@ -0,0 +1,38 @@
...
+const createArtist = (question: string, email: string) =>
+  tool({
+    description: `IMPORTANT: Always call this tool for ANY question related to creating artist.:
+    Do NOT attempt to answer questions on these topics without calling this tool first.
+
+    Example questions that MUST trigger this tool:
+    - "Create a new aritst"
+    - "I wanna create a new artist."
+
+    When in doubt, call this tool to ensure you have the most up-to-date and accurate information.`,
```
SweetmanTech: We want this function to fire in very specific contexts right?
- actual: `when in doubt` (broad trigger)
- expected: specific trigger
 
```suggestion
`,
```

SweetmanTech (approved):

PR #71, opened 2024-11-07
Tech322/fix empty reply/new artist

techeng322:


lib/tools/getCampaign.tsx
```diff
@@ -27,14 +27,23 @@ const getCampaign = (question: string, email: string, artistId: string) =>
     - "How many unique [users/fans] have participated in this campaign?"
     - "How many fans does the artist have?"
     - "What insights can we draw from the latest campaign?"
-    - "How many premium subscribers are there?"
-    - "What is the email count for users with an Apple Music account?"`,
```
SweetmanTech, line 31: Why are you removing this one?
- will this question still pass our evals?

SweetmanTech (approved):

PR #87, opened 2024-11-15
Tech322/get premium fans

techeng322:


evals/scripts/get_premium_fans_generator.py
```diff
@@ -0,0 +1,93 @@
...
+    if context:
+        context_str = json.dumps(context)
+    else:
+        context_str = "No context available."
+    content = {
+        "input": f"Context: {context_str}\n"
+            "Question: What is the total number of fans with a premium Spotify account?\n\n"
+            "Specific Focus:"
+            "- Count Queries: Respond with only the number, NOTHING ELSE.\n"
+            "Ensure your answer is data-driven, insightful, and provides clear value for understanding and acting on the fan base's behavior.",
```
SweetmanTech, line 84: This is still duplicated across both the 
- JS codebase
- Python codebase 
right?

That means future devs will need to update TWO places, or the evals will not be accurate, right?

If those previous points are true, why are you not referencing one, DRY, variable to 
- make the code easier to maintain
- prevent dev bugs in the future

SweetmanTech (approved):

PR #98, opened 2024-11-17
I add question.

techeng322:


evals/cli/oaieval.py
```diff
@@ -121,17 +121,20 @@ class OaiEvalArguments(argparse.Namespace):
     dry_run_logging: bool
 
 
-def send_slack_message(result: dict[str, Any]) -> None:
+def send_slack_message(result: dict[str, Any], question) -> None:
```
SweetmanTech: it is not always a question, right?
```suggestion
def send_slack_message(result: dict[str, Any], prompt) -> None:
```

evals/cli/oaieval.py
```diff
@@ -121,17 +121,20 @@ class OaiEvalArguments(argparse.Namespace):
...
-        # Format the message
         message_blocks = []
-        message_text = "Evaluation Results:\n"
+        message_text = f"Evaluation Results: `{question}`\n"
         for key, value in result.items():
-            message_text += f"• *{key}*: {value}\n"
-            
+            message_text += f"• *{key}*: {value}"
+            if key == "score":
+                message_text += " ✅\n" if value == 1.0 else " ❌\n"
```
SweetmanTech, line 134: I like the emojis 😊

evals/registry/evals/recoup.yaml
```diff
@@ -10,3 +10,4 @@ recoup.dev.v0:
     modelgraded_spec: closedqa
     modelgraded_spec_args:
       criteria: "correctness: The answer should be numerically correct and include appropriate units (if applicable). The solution should match exactly with the reference answer."
+      question: Recoup Chat Benchmarks
```
SweetmanTech: hmmmm, I'm not sure `prompt` or `question` mkes sense here as this text is not being passed as a prompt but rather it is a `label` for a group of evals.

What do you think of the name `label` for this variable?

SweetmanTech (approved):

PR #113, opened 2024-11-20
Tech322/video comments api

techeng322:


app/api/get_tiktok_video_comments/get_dataset_items/route.ts
```diff
@@ -0,0 +1,28 @@
+import getDefaultDataset from "@/lib/apify/getDefaultDataset";
+import getFormattedCommentsInfo from "@/lib/apify/getFormattedCommentsInfo";
+import { NextRequest } from "next/server";
+
+export async function GET(req: NextRequest) {
+  const defaultDatasetId = req.nextUrl.searchParams.get("defaultDatasetId");
```
SweetmanTech, line 6: could you please change this variable name?
- `default` infers I do not need to provide anything, and it will use this as a `fallback`
- However, this is a required param, right?
```suggestion
  const datasetId = req.nextUrl.searchParams.get("datasetId");
```

SweetmanTech (approved):

PR #114, opened 2024-11-20
Tech322/pass comments context

techeng322:


evals/scripts/create_tiktok_vi_comments_registry_data.py
```diff
@@ -6,18 +6,50 @@
...
+    response = requests.get(f"{api_endpoint}/api/get_tiktok_video_comments/get_dataset_items?datasetId={datasetId}")
+    data = response.json()
+    if len(data['data']['videos']) > 0:
+        return data['data']
+    else:
+        return None
+
 def create_tiktok_vi_comments_registry_data(ideal_key_or_value, question):
     registry_data = ""
+    defaultDatasetId = fetch_default_dataset()
```
SweetmanTech, line 28: For the evals, could you please set the `comments` to `3` to keep our costs low?
- actual: 100 comments
- expected: 3 comments

SweetmanTech: Alternatively, hard-code the `datasetId` so we do not query for new comments during evals to keep API costs low.

evals/scripts/create_tiktok_vi_comments_registry_data.py
```diff
@@ -6,18 +6,50 @@
...
-    
+
+    while True:
+        time.sleep(3)
+        data = fetch_comments(defaultDatasetId)
+        if data != None:
+            context_str = json.dumps(data)
+        else:
+            context_str = "No context available."
+        if data is not None:
```
SweetmanTech, line 38: If we `hard-code` the datasetId then we will not need to `sleep` right?

SweetmanTech (approved):

PR #163, opened 2024-12-04
add: activate event

techeng322:


components/Artists/Artist.tsx
```diff
@@ -3,13 +3,18 @@ import { ArtistRecord } from "@/types/Artist";
 import ImageWithFallback from "../ImageWithFallback";
 
 const Artist = ({ artist }: { artist: ArtistRecord }) => {
-  const { setSelectedArtist } = useArtistProvider();
+  const { setSelectedArtist, setArtistActive } = useArtistProvider();
+
+  const handleClick = (artist: ArtistRecord) => {
+    setArtistActive(true);
+    setSelectedArtist(artist);
+  };
```
sweetmantech, line 11: Now that we've remove the toggle, do we still need `both`
- `artistActive` - `boolean` as well as
- `selectedArtist` - `string`

what functionality would we lose if we deleted artistActive?

sweetmantech (approved):

PR #201, opened 2024-12-12
Techengme/myc 656 p1 subscribe unlock unlimited reports credits

techeng322:


app/api/stripe/subscription/search/route.ts
```diff
@@ -0,0 +1,29 @@
+import stripeClient from "@/lib/stripe/client";
+import { NextRequest } from "next/server";
+
+export async function GET(req: NextRequest) {
+  const accountId = req.nextUrl.searchParams.get("accountId");
```
sweetmantech, line 5: Is the goal of this API to get active subscriptions for an account?
- if yes, why not lookup subscription for the `email_address` instead of `accountId`?
- `email` can be directly queried from the stripe API so your query can be more `specific` 🎯

components/Funnels/TikTokAccountAnalysis/Segments.tsx
```diff
@@ -29,13 +33,12 @@ const Segments = () => {
...
     if (!isPrepared()) return;
-    const credits = await getCredits(userData?.id);
-    if (!credits?.remaining_credits) {
+    if (!hasCredits && !subscriptionActive) {
       setSuccessCallbackParams(new URLSearchParams({ segmentName }).toString());
       toggleModal();
       return;
     }
-    await decreaseCredits(userData?.id);
+    if (!subscriptionActive) await decreaseCredits(userData?.id);
```
sweetmantech, line 41: 💯

sweetmantech (approved):

PR #366, opened 2025-01-31
feat: fix app crash bug

techeng322:


components/Artists/Artist.tsx
```diff
@@ -32,10 +32,10 @@ const Artist = ({
         onContextMenu={handleDropDown}
       >
         <ImageWithFallback
-          src={artist.image || "https://i.imgur.com/QCdc8Ai.jpg"}
+          src={artist.artist.name || "https://i.imgur.com/QCdc8Ai.jpg"}
```
sweetmantech, line 35: seems weird to be putting `artist.name` as the src for an image 🤔

types/Artist.tsx
```diff
@@ -8,18 +7,48 @@ export type Artist = {
...
-  name: string | null;
-  organization: string | null;
+  organization: string;
+  updated_at: string;
+};
+
+export type ARTIST_INFO = {
+  account_id: string;
+  artist_id: string;
+  id: string;
```
sweetmantech, line 43: what's the difference between each of these `id`s?
- also a 4th ID in `artist.artist.id`
- are `4 diffenrent` IDs really necessary?

types/Artist.tsx
```diff
@@ -8,18 +7,48 @@ export type Artist = {
...
   updated_at: string;
+  artist: {
+    id: string;
+    name: string;
+    timestamp: number | null;
+    account_socials: Array<ACCOUNT_SOCIAL>;
+    account_info: Array<ACCOUNT_INFO>;
+  };
```
sweetmantech, line 51: why nest this `artist` object?
- it seems to cause a lot of `artist.artist` blocks of code.
- what happens if you move all attributes in `artist` to the `root` object for easier accessibility?

components/Tools/ArtistsTable.tsx
```diff
@@ -1,5 +1,5 @@
 import { useToolCallProvider } from "@/providers/ToolCallProvider";
-import { ArtistRecord } from "@/types/Artist";
+import { ARTIST_INFO } from "@/types/Artist";
```
sweetmantech, line 2: why change the name from `ArtistRecord` to `ARTIST_INFO`?
- which naming convention are we using in our app?
- please maintain `consistent naming conventions`

PR #455, opened 2025-03-13
MYC-1265: Migrate room creation to use rooms.artist_id

sidneyswift:
This PR updates the room creation logic to work with the new database structure where artist_id is in the rooms table. This enables all new chats to work correctly with the updated database schema.

**Changes:**
- Updated `createRoom` function to accept an `artist_id` parameter
- Updated room creation API to accept `artist_id` as a query parameter
- Updated `createRoomWithReport` function to include `artist_id` in the database insert
- Updated `useChat` hook to pass the selected artist's account_id as the `artist_id`
- Updated segment page to include the `artist_id` parameter in the `createRoomWithReport` call

This PR depends on the database migration that was already applied (MYC-1299).

### DIFF
--- a/lib/createRoom.tsx
+++ b/lib/createRoom.tsx
@@ -1,12 +1,18 @@
 import getAiTitle from "./getAiTitle";
 
-const createRoom = async (account_id: string, content: string) => {
+const createRoom = async (account_id: string, content: string, artist_id?: string) => {
   try {
     const title = await getAiTitle(content);
     const topic = title.replaceAll(`\"`, "");
-    const response = await fetch(
-      `/api/room/create?account_id=${account_id}&topic=${encodeURIComponent(topic)}`,
-    );
+    const url = new URL(`/api/room/create`, window.location.origin);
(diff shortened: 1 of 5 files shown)


lib/createRoom.tsx
```diff
@@ -1,12 +1,18 @@
...
 
-const createRoom = async (account_id: string, content: string) => {
+const createRoom = async (account_id: string, content: string, artist_id?: string) => {
   try {
     const title = await getAiTitle(content);
     const topic = title.replaceAll(`\"`, "");
-    const response = await fetch(
-      `/api/room/create?account_id=${account_id}&topic=${encodeURIComponent(topic)}`,
-    );
+    const url = new URL(`/api/room/create`, window.location.origin);
```
sweetmantech, line 7: Is `new URL` required here?

- before: just `fetch` from the `relative path`
- after: `new URL()`
- question: what `blocker` or `error` caused us to need to use `new URL()`?

sweetmantech (approved):

PR #473, opened 2025-03-20
MYC-1277: improve conversation context retention

sidneyswift:
This PR implements conversation context retention, allowing the AI to remember previous messages in the chat.

**Changes**
- Added getServerMessages function that retrieves up to 100 previous messages for AI context
- Renamed getInitialMessages to getClientMessages to clarify security boundaries
- Updated chat API to include previous messages when sending to the AI

**Benefits**
- AI now remembers the conversation history (up to 100 messages)
- Conversations feel more natural and coherent

**Testing**
- Chat history displays correctly in the UI
- AI responses include context from previous messages


app/api/memories/get/route.ts
```diff
@@ -1,18 +1,28 @@
...
+  }
+
   try {
+    
     const { data, error } = await supabase
       .from("memories")
       .select("*")
-      .eq("room_id", roomId);
+      .eq("room_id", roomId)
+      .order("updated_at", { ascending: true });
```
sweetmantech, line 17: Same question as the last time you opened this PR.
- What is the difference between this supabase query and `getServerMessages`?
- If no difference, Why not have this endpoint call `getServerMessages` to DRY duplicated code?

lib/supabase/getServerMessages.ts
```diff
@@ -0,0 +1,35 @@
...
+        return new HumanMessage(content.content);
+      } else if (content.role === "assistant") {
+        return new AIMessage(content.content);
+      }
+      
+      return new HumanMessage(content.content);
+    });
+    
+    
+    return langChainMessages.reverse();
```
sweetmantech, line 31: Why are you `reversing` the list?
- What ordering do you think the list is in? (latest first or earliest first)?

app/api/memories/get/route.ts
sidneyswift: I'm implementing a variation of the reviewer's suggestion.

Reviewer's suggestion: Have the API endpoint call getServerMessages to avoid duplicate code.

My plan: Create a shared helper function in getServerMessages.ts that both the API and getServerMessages itself can use.

The difference is subtle but important:
- Direct implementation of reviewer's suggestion: API endpoint would call a function that does extra processing it doesn't need

- My plan: Extract just the database query part for both to use, keeping separate processing logic
This approach addresses the reviewer's main concern (duplicate query code) while avoiding potential issues. 

Both implementations use the same database query logic, just packaged differently than the reviewer suggested.

lib/supabase/getServerMessages.ts
sidneyswift: We query with descending order (newest first) so we get the most recent messages when applying the limit, then reverse the results to get them in chronological order (oldest first) so LLM reads messages in order of conversation flow.

lib/supabase/getServerMessages.ts
```diff
@@ -0,0 +1,50 @@
...
+    .order("updated_at", { ascending });
+    
+  if (limit) {
+    query = query.limit(limit);
+  }
+  
+  return query;
+}
+
+export async function getServerMessages(roomId: string, limit = 100): Promise<BaseMessage[]> {
```
sweetmantech, line 25: why do we need BOTH
- queryMemories
- getServerMessages

in my opinion, only one is required.

sweetmantech (approved):

PR #487, opened 2025-04-02
MYC-1402-p0-recent-chat-organization

sidneyswift:
MYC-1282: Filter and Sort Recent Chats

This PR implements the first phase of the chat management improvements by:
Filtering the Recent Chats list to only show conversations for the currently selected artist (instead of all account chats)
Sorting conversations by recency, with the most recent chats appearing at the top of the list

Changes Made:
Updated components/Sidebar/RecentChats.tsx to use the filtered conversations array instead of allConversations
Modified hooks/useConversations.tsx to sort the filtered conversations by updated_at timestamp in descending order
These changes improve the user experience by showing only relevant chats for the current artist and prioritizing recent conversations, making navigation more intuitive and efficient.
The second phase will implement the rename and delete functionality, which will be added in a subsequent PR.

### DIFF
--- a/hooks/useConversations.tsx
+++ b/hooks/useConversations.tsx
@@ -33,7 +33,14 @@ const useConversations = () => {
       (item: Conversation | ArtistAgent) =>
         'artist_id' in item && item.artist_id === selectedArtist?.account_id
     );
-    return filtered;
+    
+    // Sort by updated_at in descending order (most recent first)
+    return filtered.sort((a, b) => {
+      if ('updated_at' in a && 'updated_at' in b) {
+        return new Date(b.updated_at).getTime() - new Date(a.updated_at).getTime();
+      }
+      return 0;
+    });
   }, [selectedArtist, allConversations]);
 
   const fetchConversations = async () => {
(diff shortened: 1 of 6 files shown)


hooks/useConversations.tsx
```diff
@@ -33,7 +33,14 @@ const useConversations = () => {
...
     );
-    return filtered;
+    
+    // Sort by updated_at in descending order (most recent first)
+    return filtered.sort((a, b) => {
+      if ('updated_at' in a && 'updated_at' in b) {
+        return new Date(b.updated_at).getTime() - new Date(a.updated_at).getTime();
+      }
+      return 0;
+    });
```
sweetmantech, line 43: How could we handle ordering at the most upstream point so the data is in the correct order in any part of the app?
- actual: ordering in `hook` useConversations (downstream)
- suggested: ordering in `supabase query` (source) app/api/room/get/route.tsx

My suggestion should result in
- less code changes
- sorted chat across the app (not just useConversations)
Screenshot 2025-04-03 at 1 02 38 PM

PR #553, opened 2025-05-05
Veerbalsingh369/myc 1744 p1 show recent chat room on creation

veerbal1:
Before: 
- When we create a new conversation, it does not refresh the conversation list (on the left sidebar). Users have to manually refresh the page to see the latest conversation item in the sidebar. 

After this PR: 
- When we create a new conversation, it will automatically trigger a refetch of the conversation list, showing the updated list. No need to refresh the page.
- Highllight (selected) newly created conversation in sidebar


hooks/useVercelChat.ts
```diff
@@ -28,6 +30,8 @@ export function useVercelChat({ id, initialMessages }: UseVercelChatProps) {
   const userId = userData?.id;
   const artistId = selectedArtist?.account_id;
   const [hasChatApiError, setHasChatApiError] = useState(false);
+  const messagesRef = useRef<ChatMessage[]>([]);
```
sweetmantech, line 33: why are you creating a new `ref`?
`messages` already exists from `useChat` from the `ai` package right?
why do you need a duplicate ref?

components/Sidebar/RecentChats.tsx
```diff
@@ -25,8 +25,31 @@ const RecentChats = ({ toggleModal }: { toggleModal: () => void }) => {
...
+        // Handle params safely with type checking
+        const roomId = typeof params?.roomId === 'string' ? params.roomId : null;
+        const agentId = typeof params?.agentId === 'string' ? params.agentId : null;
+        setActiveChatId(roomId || agentId || null);
+      }
+    };
+
+    // Update on initial render and when dependencies change
+    updateActiveChatId();
+  }, [params, conversations]);
```
sweetmantech, line 52: Is this how the vercel `chat sdk` handles activeConversation highlights in recent chats?
- personally, this feels like a `code smell` and could be simplified.
- `history.replace` is used by the vercel chat sdk.
- please reference the official vercel chat sdk and link the reference you're following here so I can verify we are following the `top tier` reference 🔗

hooks/useVercelChat.ts
veerbal1: Using direct messages inside onFinish is not an ideal approach because messages inside onFinish will always remain behind the real updated value and are dependent on the useChat re-rendering. onFinish is a function inside a function (useChat), which acts as a closure. And inside the function, it gets the value at the time of initialization. In our case, messages inside onFinish will always get a new value when useChat somehow re-renders, which reinitializes the onFinish function. 

So messagesRef.current will always have an updated value (as messages change), irrespective of whether onFinish gets called or not.

I checked Vercel's Chat SDK code. They are not even using messages, messagesRef. etc. But there is a big problem in their code. Correct me if  I am wrong

Screenshot 2025-05-12 at 11 11 40 PM

They are triggering the chat history invalidation endpoint on every onFinish call, and onFinish gets called after every message stream Finish. 

It means fetching chat history again and again on every message stream. I don't think it is a planned or a smart approach. 

And in our case we are refetching the list just when the 2nd message stream and get updated in the database, just one time. No unnecessary calls.

If you want then I can write similar code as Vercel Chat SDK, then there will no need to write messageRef. code.

components/Sidebar/RecentChats.tsx
veerbal1: There is a bug in Vercel Chat SDK that we just fixed in Recoup in this PR. 

To reproduce - go to https://demo.chat-sdk.dev/
- Log in and start a new chat
- ask some questions

See the bug in the left sidebar - 
- A new chat will be created (that's ok)
- It is not selected (new current chat)
- Click on the newly added chat, it gets selected when we click on it.

why? 
because they are using history.replace, it is a good approach to start a chat on the same page without navigating to the chat/{chatId} page. (instant chat)

Screenshot 2025-05-12 at 11 33 19 PM

history.replace silently replace URL in browser outside of Nextjs and it does not cause rerendering, because it does not cause re-rendering, it means useParams hook also does not get updated. it has no idea what has changed inside the URL, so we have no updated active chat id and we are not able to highlight the current active chat nav link.

How I am fixing it?. 
Because we can't rely on useParams or any path hook, I am relying on the conversation list array because it gets updated by the onFinish function inside the useChat hook. (Check below code)

When the conversation list array changes, it calls useEffect, which checks what the current params are in the URL, and it changes the internal activeChatId state, which triggers conversation list re-render, and we automatically getthe  selected chat item.

sweetmantech (approved):

PR #560, opened 2025-05-06
Veerbalsingh369/myc 1535 p1 add file upload

veerbal1:


app/api/files/upload/route.ts
sweetmantech, line 1: why are you creatign a new upload route instead of using the existing `/api/upload` route?

components/VercelChat/ChatInput.tsx
```diff
@@ -22,44 +26,61 @@ export function ChatInput({
...
+          <button
+            type="submit"
+            className={cn(
+              "size-8 flex flex-row justify-center items-center dark:bg-zinc-100 bg-zinc-900 dark:text-zinc-900 text-zinc-100 p-1.5 rounded-full hover:bg-zinc-800 dark:hover:bg-zinc-300 hover:scale-105 active:scale-95 transition-all",
+              {
+                "dark:bg-zinc-200 dark:text-zinc-500 cursor-not-allowed opacity-50":
+                  isGeneratingResponse || input === "" || isDisabled || hasPendingUploads,
+              }
+            )}
+            disabled={isGeneratingResponse || input === "" || isDisabled || hasPendingUploads}
```
sweetmantech, line 78: this logic is repeated 2x 
- please follow the DRY principle by creating a shared variable which is easier to maintain.

components/VercelChat/PureAttachmentsButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const fileInputRef = useRef<HTMLInputElement>(null);
+    const MAX_FILES = 10;
+
+    const uploadFile = async (file: File) => {
+        // Only allow image files for now
+        const allowedTypes = ['image/jpeg', 'image/png', 'image/gif', 'image/webp'];
+        if (!allowedTypes.includes(file.type)) {
+            console.error('File type not supported:', file.type);
+            return;
+        }
```
sweetmantech, line 18: if you're filtering this in the `input` why do you need this logic here for `allowedTypes`?
- can this code be removed? if no, why not?

components/VercelChat/PureAttachmentsButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+        try {
+            // Upload the file to Arweave
+            const formData = new FormData();
+            formData.append('file', file);
+
+            const response = await fetch('/api/files/upload', {
+                method: 'POST',
+                body: formData,
+            });
```
sweetmantech, line 39: why aren't you using the existing `/api/upload` endpoint here?

components/VercelChat/PureAttachmentsButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+            console.warn(`Only the first ${MAX_FILES} files will be uploaded`);
+        }
+        
+        filesToUpload.forEach(uploadFile);
+        
+        // Reset the input
+        if (fileInputRef.current) {
+            fileInputRef.current.value = '';
+        }
+    };
```
sweetmantech, line 95: the first 100 lines of this React component is JS / state code.
- please keep component files to only `HTML` / JSX tags
- please move the first 100 lines of this file to a `hook` (ex useFileAttachments)
- all files should have `less than 100 lines of code`

components/VercelChat/chat.tsx
```diff
@@ -45,7 +45,7 @@ function ChatContent({ reportId, id }: { reportId?: string; id: string }) {
     input,
     setMessages,
     reload,
-  } = useVercelChatContext();
+  } = useVercelChatContext()
```
sweetmantech, line 48: is this change adding any real functionality?
if no, please remove it
`git checkout origin/main components/VercelChat/chat.tsx`

components/VercelChat/preview-attachment.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <div className="flex items-center justify-center text-zinc-500">
+              <FileIcon size={24} />
+              <span className="text-xs ml-1">.pdf</span>
+            </div>
+          ) : (
+            // Generic file icon
+            <div className="flex items-center justify-center text-zinc-500">
+              <FileIcon size={24} />
+            </div>
+          )
```
sweetmantech, line 36: only images are supported right?
Why do you have so much extra logic if you're only supporting images?

hooks/useVercelChat.ts
```diff
@@ -10,18 +10,24 @@ import { clientDeleteTrailingMessages } from "@/lib/messages/clientDeleteTrailin
 import { generateUUID } from "@/lib/generateUUID";
 import { usePrivy } from "@privy-io/react-auth";
 import { useConversationsProvider } from "@/providers/ConversationsProvider";
+import { Attachment } from "@ai-sdk/ui-utils";
```
sweetmantech, line 13: nice use of existing SDK types ✅

hooks/useVercelChat.ts
```diff
@@ -10,18 +10,24 @@ import { clientDeleteTrailingMessages } from "@/lib/messages/clientDeleteTrailin
...
 /**
  * A hook that provides all chat functionality for the Vercel Chat component
  * Combines useChat, and useMessageLoader
  * Accesses user and artist data directly from providers
  */
-export function useVercelChat({ id, initialMessages }: UseVercelChatProps) {
+export function useVercelChat({ 
+  id, 
+  initialMessages,
+  uploadedAttachments = [] // Default to empty array
```
sweetmantech, line 29: in what case is `uploadedAttachments` needing to be passed in the initialization of the `useVercelChat` hook?
- personally, I think the `uploadedAttachments` param can be `removed` from the initialization of `useVercelChat` 🚮

hooks/useVercelChat.ts
```diff
@@ -50,6 +56,7 @@ export function useVercelChat({ id, initialMessages }: UseVercelChatProps) {
       accountId: userId,
       email: userData?.email,
     },
+    // initialMessages,
```
sweetmantech, line 59: this should have been removed from `main`
are you sure you have the latest changes from `main`?
`git pull origin main`
- overall: delete unused code.

hooks/useVercelChat.ts
```diff
@@ -117,12 +124,20 @@ export function useVercelChat({ id, initialMessages }: UseVercelChatProps) {
     window.history.replaceState({}, "", `/chat/${id}`);
   };
 
-  const handleSendMessage = async () => {
+  const handleSendMessage = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
```
sweetmantech, line 128: why do you need to `preventDefault`?

providers/VercelChatProvider.tsx
```diff
@@ -10,13 +12,20 @@ interface VercelChatContextType {
   isLoading: boolean;
   hasError: boolean;
   isGeneratingResponse: boolean;
-  handleSendMessage: () => Promise<void>;
+  handleSendMessage: (event: React.FormEvent<HTMLFormElement>) => Promise<void>;
```
sweetmantech, line 15: why do you need the `event` as an additional param in `handleSendMessage`?
- can this be removed? 🚮

providers/VercelChatProvider.tsx
```diff
@@ -10,13 +12,20 @@ interface VercelChatContextType {
...
   stop: UseChatHelpers["stop"];
   setInput: UseChatHelpers["setInput"];
   input: UseChatHelpers["input"];
   setMessages: UseChatHelpers["setMessages"];
   reload: UseChatHelpers["reload"];
   append: UseChatHelpers["append"];
+  attachments: Attachment[];
+  pendingAttachments: Attachment[];
+  uploadedAttachments: Attachment[];
```
sweetmantech, line 24: what's the difference between `attachments` and `uploadedAttachments`?

providers/VercelChatProvider.tsx
```diff
@@ -39,6 +48,17 @@ export function VercelChatProvider({
...
+  // Use the useAttachments hook to get attachment state and functions
+  const {
+    attachments,
+    pendingAttachments,
+    uploadedAttachments,
+    setAttachments,
+    removeAttachment,
+    clearAttachments,
+    hasPendingUploads,
+  } = useAttachments();
```
sweetmantech, line 60: why not move `useAttachments` import inside `useVercelChat` so you don't need to pass additional params?

providers/VercelChatProvider.tsx
```diff
@@ -53,7 +73,19 @@ export function VercelChatProvider({
...
+    uploadedAttachments // Pass attachments to useVercelChat
+  });
+  
+  // When a message is sent successfully, clear the attachments
+  const handleSendMessageWithClear = async (event: React.FormEvent<HTMLFormElement>) => {
+    await handleSendMessage(event);
+    
+    // Clear attachments after sending
+    clearAttachments();
+  };
```
sweetmantech, line 88: why not just call this in `onFinish` callback in `useChat` hook?

components/VercelChat/PureAttachmentsButton.tsx
veerbal1: This is a double check in JS code in case the user tries to bypass the allowed files in the file input through the browser web inspector.

The user can easily bypass it by removing the accept attribute from the file input. If we do not have a check in JS, then this input will also accepts video files, almost anything, and upload tothe  file storage database

components/VercelChat/preview-attachment.tsx
veerbal1: It is a extension, in case in future we other files than just images.

Removing it for now

hooks/useVercelChat.ts
veerbal1: `uploadedAttachments ` is an optional parameter for useVercelChat, and it is not needed for its initialization.

We have a dedicated hook called `useAttachments`. Its job is to manage uploaded files, tracking which file is still uploading or has been uploaded. to remove the selected file, etc.

useVercelChat's job is to manage our chat, communication with the LLM API

These 2 hooks are being used in a global provider `VercelChatProvider`

`useAttachments` hook is just passing information about the selected and uploaded files to `useVercelChat` hook so that it can attach files with handleSubmit function

providers/VercelChatProvider.tsx
veerbal1: It is for the separation of concerns and it is ideal to use inside `VercelChatProvider` (reason explained at bottom)

`useAttachments` track uploaded, pending files, and manage file removal.

`useVercelChat ` is to manage communication with llm API.

`useVercelChat` just needs information about selected and uploaded files, and only when they are dropped by the user.

+ 

useAttachments should be placed inside `VercelChatProvider` as its values will be used to show uploading UI elements for files being uploaded. it is out of `useVercelChat`'s responsibility.

providers/VercelChatProvider.tsx
veerbal1: Nope, 
it will keep showing files above the input even after the user submits the message, and it will be showing images even when the AI message is streaming. Suppose AI takes 30 seconds to answer (reasoning + response), the user will still be looking at the files above the input and it will only disappear when message stream end. 

It will be a bad UX and we just attach files with user's message and submit. then we don't have any use of files after the submission.

providers/VercelChatProvider.tsx
veerbal1: `handleSendMessage` is attached to the form's onSubmit handler, which passes a form event, and TypeScript gives an error if we don't allow any event in the handleSendMessage function

+ Apart from this, I was having an issue with the attached files' information not being visible on the backend, but that worked when I passed form event.

hooks/useVercelChat.ts
veerbal1: `handleSendMessage` is attached to the form's onSubmit handler. It triggers a GET request on submit. (old ways of submitting data through an HTML form, method='GET')

As we are managing everything on the client side (React app) so we are using event.preventDefault() function to prevent form's default behavior (from making a GET request, causing page reload with URL params)

so that's the reason. I can test it by removing, see how it behave

components/VercelChat/PureAttachmentsButton.tsx
sweetmantech: that sounds like an extreme edge case yeah? in that case, what would happen? why not let the hacket try to send the custom fily type?

providers/VercelChatProvider.tsx
sweetmantech: good call 👍

sweetmantech (approved):

sweetmantech (approved):

PR #587, opened 2025-05-16
Veerbalsingh369/myc 1901 generate mermaid chart copy button

veerbal1:


components/Chat/mermaid/MermaidDiagram.tsx
```diff
@@ -40,15 +44,40 @@ const MermaidDiagramInternal: React.FC<MermaidDiagramProps> = ({ chart, id }) =>
...
+                aria-label="Download image"
+              >
+                <Download className="h-4 w-4 text-gray-700" />
+              </Button>
+            </TooltipTrigger>
+            <TooltipContent>
+              <p>Download</p>
+            </TooltipContent>
+          </Tooltip>
+        </TooltipProvider>
```
sweetmantech, line 75: Is this the same code you just wrote for the image download button?
what would happen if we followed the DRY principle to create a shared <DownloadButton /> component imported by both components?

hooks/useMermaidDownload.ts
sweetmantech, line 1: why is this a `hook` rather than simply a `lib`?

sweetmantech (approved):

PR #616, opened 2025-05-21
Disabling 'create new artist' button when a new artist is being created

veerbal1:


hooks/useArtists.tsx
```diff
@@ -51,14 +52,18 @@ const useArtists = () => {
         ]
       : sortArtistsAlphabetically(artists);
 
+  const handleSetIsArtistBeingCreated = useCallback((value: boolean) => {
+    setIsArtistBeingCreated(value);
+  }, []);
```
sweetmantech, line 57: why do you need this `handle` function with `useCallback`?
- what happens if you directly export the `setIsArtistBeingCreated` function instead of wrapping it?

hooks/useClickChat.tsx
```diff
@@ -1,16 +1,20 @@
...
+  const { setIsArtistBeingCreated } = useArtistProvider();
+
 
   const handleClick = (conversation: any, toggleModal: () => void) => {
     if (isMobile) toggleModal();
     if (conversation?.agentId) {
       push(`/funnels/${conversation?.type}/${conversation.agentId}`);
       return;
     }
+    setIsArtistBeingCreated(false);
```
sweetmantech, line 17: I'm not liking that you need to use `setIsArtistBeingCreated` across multiple different files.
- I'd prefer the `isArtistBeingCreated` state be `self-contained` without needing to manage the state in multiple, unrelated, components.

Would the files being changed be `reduced` if you
- make a new hook `useCreateArtist`
- add a useEffect to setIsArtistBeingCreated to `false` on route change
- import the `useCreateArtist` hook in `useArtists` so it can be accessed globally in the provider

PR #622, opened 2025-05-22
Veerbalsingh369/myc 1978 p0 create artist button multiple clicks

veerbal1:


providers/VercelChatProvider.tsx
```diff
@@ -58,7 +67,8 @@ export function VercelChatProvider({
     clearAttachments,
     hasPendingUploads,
   } = useAttachments();
-  
+  const { updateChatState } = useArtistProvider();
+
```
veerbal1, line 71: Focus Point 1

PR #629, opened 2025-05-25
Sidney/myc 1915 p1 agents tab redesign starter agent library integration

sidneyswift:
This PR redesigns the “Agents” tab to load agent cards dynamically from agents.json. Only active agents are shown, tag filters are generated from the metadata, and the “Fan Segmentation Analysis” card always appears first and uses the legacy funnel logic. The UI is now cleaner and easier to maintain. Unrelated files have been removed from the PR.

**How to test:**
- Open the “Agents” tab.
- Check that agent cards and tags match agents.json.
- Confirm the “Fan Segmentation Analysis” card is first and works as before.
- Try different tag filters and verify the grid updates.
- Add a new agent or tag to agents.json and see it appear in the UI.


app/chat/page.tsx
```diff
@@ -10,7 +10,10 @@ interface ChatPageProps {
 
 export default async function ChatPage({ searchParams }: ChatPageProps) {
   const id = generateUUID();
-  const initialMessage = (await searchParams)?.q as string;
+  const params = await searchParams;
+  const initialMessage =
+    (params?.starterPrompt as string) ||
+    (params?.q as string);
```
sweetmantech, line 16: how is `starterPrompt` different from `q`?

why not use the existing `q` param instead of creating a new `searchParam`?

components/Agents/Agents.tsx
```diff
@@ -18,73 +92,88 @@ const Agents = () => {
...
+            >
+              <div className="text-lg mb-1 font-medium">
+                {agent.title}
+              </div>
+              <div className="text-gray-400 text-sm">{agent.description}</div>
+            </button>
+          ))}
+        </div>
+      )}
+      <div className="mb-16" />
```
sweetmantech, line 176: file is too long.
- required: files are `less than 100 lines of code`
- actual: file is `179 lines` (too long)

recommended: 
- new hook `useRecipes`
- new component: `<RecipeButton />`
- new component: `<RecipeTags />`
- etc

public/agents.json
```diff
@@ -0,0 +1,170 @@
+[
+  {
+    "title": "Fan Segmentation Analysis",
+    "description": "Get a 360° breakdown of your fanbase and audience segments.",
+    "prompt": "",
+    "tags": ["Recommended", "Wrapped", "Analysis"],
+    "status": "active",
+    "isActive": true
```
sweetmantech, line 8: why do we need `isActive`?
- why not just `delete` the recipe if it is not active?
- in what condition do we want a `non-active` recipe?

sweetmantech: I recommend removing both
- `status`
- `isActive`

keep these:
- `title`
- `description`
- `prompt`
- `tags`

app/chat/page.tsx
sidneyswift: Simplified: Now using only the q param for all initial messages.

app/chat/page.tsx
```diff
@@ -10,7 +10,9 @@ interface ChatPageProps {
 
 export default async function ChatPage({ searchParams }: ChatPageProps) {
   const id = generateUUID();
-  const initialMessage = (await searchParams)?.q as string;
+  const params = await searchParams;
+  // Use only 'q' for the initial message, regardless of source (agent or user).
+  const initialMessage = params?.q as string;
```
sweetmantech, line 15: are these changes necessary?
- what `problem` are they solving or `value` is being added with `extra lines of code`?

I suggest `git checkout origin/main app/chat/page.tsx`

sweetmantech (approved):

PR #690, opened 2025-06-02
Created Youtube Tool call function that fetches Youtube details

veerbal1:


lib/tools/getYouTubeChannelInfo.ts
```diff
@@ -0,0 +1,118 @@
...
+});
+
+const getYouTubeChannelInfoTool = tool({
+  description:
+    "Get YouTube channel information for a specific account. " +
+    "This tool automatically checks authentication status and either returns channel data or authentication instructions. " +
+    "No need to check authentication separately - this tool handles everything internally. " +
+    "Returns comprehensive channel data including statistics, thumbnails, and branding if authenticated, " +
+    "or clear authentication instructions if tokens are missing/expired. " +
+    "IMPORTANT: This tool requires the artist_id parameter. If you don't know the artist_id, ask the user or use the current artist's artist_id.",
```
sweetmantech, line 31: does this match the variable name in the system prompt?
- I recommend explicitly stating this variable should come from the system prompt. be sure the spelling is precisely the same.

lib/tools/getYouTubeChannelInfo.ts
```diff
@@ -0,0 +1,118 @@
...
+            defaultLanguage: channel.branding?.defaultLanguage || null,
+          },
+          
+          // Authentication metadata
+          authentication: {
+            tokenCreatedAt: tokenValidation.tokens!.created_at,
+            tokenExpiresAt: tokenValidation.tokens!.expires_at,
+          },
+        }
+      });
```
sweetmantech, line 97: why are you destructuring the response?
- code would be more efficient to simply return `{...channel, ... tokenValidation}` right?

sweetmantech (approved):

PR #873, opened 2025-07-16
Moved StandaloneYoutubeComponent to a dedicated file

veerbal1:


components/ArtistSetting/StandaloneYoutubeComponent/ChannelInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+import formatFollowerCount from "@/lib/utils/formatFollowerCount";
+import { useQuery } from "@tanstack/react-query";
+import { Youtube } from "lucide-react";
+
+const ChannelInfo = ({ dense, artistAccountId }: { dense?: boolean; artistAccountId: string }) => {
+  const { data, isLoading } = useQuery({
+    queryKey: ["youtube-channel-info", artistAccountId],
+    queryFn: () =>
+      fetch(
+        `/api/youtube/channel-info?artist_account_id=${artistAccountId}`
```
sweetmantech, line 12: Do we need all of these?
- /api/youtube/token
- /api/youtube/status
- /api/youtube/channel
IMO, get rid of all of these and keep only `/api/youtube/channel`

principle
```
less is more
simple is best
small is beautiful
```

PR #994, opened 2025-08-14
Veerbalsingh369/myc 2717 default progress bar around pfp

veerbal1:


veerbal1:
It has been completed by Sweetman

PR #999, opened 2025-08-15
Veerbalsingh369/myc 2723 refactor userprofilebutton component

veerbal1:


components/Sidebar/UserProfileButton.tsx
```diff
@@ -49,29 +38,7 @@ const UserProfileButton = () => {
...
-            </DropdownMenuItem>
-          </DropdownMenuGroup>
-          <DropdownMenuSeparator />
-          <DropdownMenuItem className="cursor-pointer" onClick={signOut}>
-            <IconLogout />
-            Log out
-          </DropdownMenuItem>
-        </DropdownMenuContent>
-      </DropdownMenu>
+      <UserProfileDropdown onProfileClick={toggleModal} onSignOut={signOut} />
```
sweetmantech, line 41: Why pass these as props?
- Wouldn't the code be simpler if you just imported both of these from the `useUserProvider` directly within `UserProfileDropdown`?

PR #1110, opened 2025-09-10
Veerbalsingh369/myc 2835 p0 file storage and organization implement file storage

veerbal1:


supabase/migrations/20250910120000_create_files_table.sql
```diff
@@ -0,0 +1,63 @@
...
+    "artist_account_id" uuid not null,
+    "storage_key" text not null,
+    "file_name" text not null,
+    "mime_type" text,
+    "size_bytes" bigint,
+    "description" text,
+    "tags" text[] default '{}'::text[],
+    "shared_with_artist_account_ids" uuid[] default '{}'::uuid[],
+    "shared_with_account_ids" uuid[] default '{}'::uuid[],
+    "shared_with_all_artist_members" boolean not null default false,
```
sweetmantech, line 13: How will these `shared` fields be used?
- arrays are hard to query on.
- typically, we'd either create multiple entries or create a new table which is easier to query this.

sweetmantech: For example, how would I query all files shared with my account / artist?

sweetmantech: Additionally, none of these `shared` fields are foreign keys to the `accounts` table.

PR #1142, opened 2025-09-16
feat: create file_access table for managing per-file access grants

veerbal1:
- Introduced a new table "file_access" to store access grants for files, including fields for user and organization IDs, access scope, and timestamps for expiration and revocation.
- Added foreign key constraints to link with existing "files" and "accounts" tables.
- Implemented row-level security and various indexes to optimize access queries.
- Defined access grants for different user roles: anon, authenticated, and service_role.

### DIFF
--- a/supabase/migrations/20250916130500_create_file_access_table.sql
+++ b/supabase/migrations/20250916130500_create_file_access_table.sql
@@ -1,0 +1,87 @@
+-- Create table to store per-file access grants
+
+-- Simple scope enum for access level
+do $$ begin
+  if not exists (select 1 from pg_type where typname = 'file_access_scope') then
+    create type file_access_scope as enum ('read_only','admin');
+  end if;
+end $$;
+
+create table if not exists "public"."file_access" (
+    "id" uuid not null default gen_random_uuid(),
+    "file_id" uuid not null,
+    "artist_account_id" uuid,
+    "user_id" uuid,
+    "organization_id" uuid,
+    "granted_by" uuid,
+    "scope" file_access_scope not null default 'read_only',
+    "expires_at" timestamp with time zone,
+    "revoked_at" timestamp with time zone,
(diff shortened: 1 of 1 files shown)


supabase/migrations/20250916130500_create_file_access_table.sql
```diff
@@ -0,0 +1,87 @@
...
+    create type file_access_scope as enum ('read_only','admin');
+  end if;
+end $$;
+
+create table if not exists "public"."file_access" (
+    "id" uuid not null default gen_random_uuid(),
+    "file_id" uuid not null,
+    "artist_account_id" uuid,
+    "user_id" uuid,
+    "organization_id" uuid,
```
sweetmantech, line 15: Why do you need both `user_id` and `organization_id`?
- Files should not need to have any awareness of `organization`.
- It seems you are making this table `too complex`. Please remove the excessive `organization_id` column.

supabase/migrations/20250916130500_create_file_access_table.sql
```diff
@@ -0,0 +1,87 @@
...
+    "granted_by" uuid,
```
sweetmantech, line 16: why does `granted_by` matter?
- This feels like another excessive column.
- Please remove all excessive columns and only ship the minimum columns required to achieve the base functionality.

supabase/migrations/20250916130500_create_file_access_table.sql
```diff
@@ -0,0 +1,87 @@
...
+    "scope" file_access_scope not null default 'read_only',
+    "expires_at" timestamp with time zone,
+    "revoked_at" timestamp with time zone,
```
sweetmantech, line 19: both of these are irrelevant. Please remove them or give a clear reason why this is required in the v1 functionality of the files tab.

PR #1629, opened 2026-04-02
refactor: integrate Privy access token handling in delete and update

pradipthaadhi:
* Updated useDeleteScheduledAction and useUpdateScheduledAction hooks to directly use getAccessToken from usePrivy for authentication.
* Enhanced error handling to prompt users to sign in if the access token is not available.
* Modified deleteTask and updateTask functions to accept the access token as a parameter for API calls, ensuring secure and authenticated requests.

## Bug Fixes

* Task deletion and updates now require user authentication. Users will see a sign-in prompt if attempting to delete or update tasks while not signed in, with improved error messaging to guide the authentication flow.

### DIFF
--- a/lib/tasks/deleteTask.ts
+++ b/lib/tasks/deleteTask.ts
@@ -1,4 +1,3 @@
-import { ScheduledAction } from "@/components/VercelChat/types";
 import { TASKS_API_URL } from "@/lib/consts";
 
 export interface DeleteTaskParams {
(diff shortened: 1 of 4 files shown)


lib/tasks/deleteTask.ts
```diff
@@ -26,39 +25,49 @@ async function deleteTaskFromDatabase(taskId: string): Promise<void> {
...
+    body: JSON.stringify({
+      id: params.id,
+    }),
+  });
 
-      if (isScheduleNotFoundError(errorText)) {
-        await deleteTaskFromDatabase(params.id);
-        return;
-      }
+  const responseText = await response.text();
```
sweetmantech, line 52: KISS - Why are you changing the error handling in this lib?

PR #1664, opened 2026-04-09
refactor: align file mentions with sandbox files flow

arpitgupta1214:
## Summary
This PR has been repurposed after the dedicated files-list API migration was rolled back.

- Keep mentions aligned with the same sandbox-backed source used by the Files page (`/api/sandboxes` + `/api/sandboxes/file`)
- Remove legacy/unused chat-side file-listing path that depended on `/api/files/list`
- Restore mention suggestion image thumbnails using sandbox file content
- Consolidate duplicate preview/content logic into a single reusable hook

## Why this changed
The original plan introduced a separate files-list endpoint for mentions, but the Files section already relies on sandbox APIs. To avoid parallel data paths and closed API/docs workstreams, mentions now reuse the existing sandbox file flow.

## Key Changes
- Mentions source
  - `useArtistFilesForMentions` now reads from sandbox filetree (`useSandboxes`) rather than files table listing
- Mention attachments + context in chat
  - Mentioned files are resolved via sandbox file API (`getFileContents`)
  - Images/PDFs are attached as file parts (data URLs), text file contents are prepended as context
- Suggestion thumbnails
  - Thumbnail rendering restored in `SuggestionItem`
  - Uses sandbox file content fetch path and local caching behavior through shared hook
- Hook consolidation
  - `useSandboxFileContent` refactored to support both:
    - select-driven sandbox file preview usage
    - path-driven usage for mention thumbnails
  - Removed duplicate `useSandboxFilePreview`
- Cleanup
  - Removed unused legacy route/hook path in chat:
    - `/app/api/files/list/route.ts`
    - `hooks/useFilesManager.ts`
    - `hooks/useBatchSignedUrls.ts`

## Verification
- Targeted lint checks on all touched files in this PR pass.
- Full typecheck still has unrelated pre-existing test typing errors in `lib/emails/__tests__/extractSendEmailResults.test.ts`.

## Notes
- Backend `/api/sandboxes/file` returns `{ status, content, encoding? }`.
- `imageUrl` is derived client-side in `getFileContents` when `encoding === "base64"`.

## Summary by CodeRabbit

* **New Features**
  * Mentions now read directly from the sandbox file tree with on-demand file content and MIME detection.
  * Image previews are loaded from sandbox content when available.

* **Bug Fixes**
  * More reliable suggestion rendering and preview fallbacks for non-image files.

* **Refactor**
  * Removed legacy signed-URL batching and moved attachment/context handling to sandbox-based retrieval.

* **Chores**
  * Cleaned up obsolete file-listing and file-manager code paths.


sweetmantech:
## Change Request

@arpitgupta1214 — after reviewing this PR and the related docs PR (recoupable/docs#122), I think we should take a different approach here.

### Problem with the current approach

The `@` mention file suggestions (`useArtistFilesForMentions`) currently fetch from a local `/api/files/list` route. This PR migrates that to a dedicated `/api/files` endpoint on the API — but that introduces a second endpoint for file listing when we already have one.

### What we already have

The **Files page** (`app/files/page.tsx`) already pulls file data from `/api/sandboxes` via `useSandboxes`. That endpoint returns the same file tree data the `@` mention feature needs.

### Requested change

Instead of migrating to a new `/api/files` endpoint, refactor `useArtistFilesForMentions` to consume the sandbox data from `useSandboxes` (or share its query). This way:

- We eliminate the need for a separate `/api/files` endpoint entirely
- Both the Files page and `@` mentions use the same data source (`/api/sandboxes`)
- We can close the docs PR (recoupable/docs#122) since `/api/files` won't be needed
- Less surface area to maintain

Let me know if you have questions or see any reason the sandbox data wouldn't work for the `@` mentions use case.

hooks/useArtistFilesForMentions.ts
```diff
@@ -1,35 +1,51 @@
 "use client";
 
 import { useMemo } from "react";
-import getRelativeStoragePath from "@/utils/getRelativeStoragePath";
-import useFilesManager, { ListedFileRow } from "@/hooks/useFilesManager";
+import useSandboxes from "@/hooks/useSandboxes";
+import getMimeFromPath from "@/lib/files/getMimeFromPath";
+import type { FileNode } from "@/lib/sandboxes/parseFileTree";
+
+const WORKSPACE_ORGS_PREFIX = ".openclaw/workspace/orgs/";
```
sweetmantech: DRY principle 
- actual: this prefix is duplicated across `/files` page and `file mentions` in the chat.
- required: shared const imported across both usages.

lib/sandboxes/workspaceOrgsPath.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Canonical path inside a sandbox where per-org/artist workspace files live.
+ * Shared by the `/files` page filetree and the `@`-mention file picker so both
+ * stay in lockstep if the sandbox layout changes.
+ */
+export const WORKSPACE_ORGS_PATH = ".openclaw/workspace/orgs";
+export const WORKSPACE_ORGS_PREFIX = `${WORKSPACE_ORGS_PATH}/`;
```
sweetmantech: DRY 
- actual: exporting WORKSPACE_ORGS_PATH and WORKSPACE_ORGS_PREFIX
- required: only export WORKSPACE_ORGS_PATH. If trailing `/` is needed, update the caller to include it.

sweetmantech (approved):

PR #1758, opened 2026-05-29
feat(chat): session-scoped sidebar rename + archive-on-delete

arpitgupta1214:
Stacked on chat#1756. Two related changes to the sidebar's chat actions:

### Rename — session-scoped PATCH
The rename flow was still hitting the legacy `PATCH /api/chats` shape (`{ chatId, topic }`), which only touched the `rooms` table. Repoint at `PATCH /api/sessions/{sid}/chats/{cid}` with `{ title }` so the canonical chat row is updated.

- `updateChat({ sessionId, chatId, title })` → `PATCH /api/sessions/{sid}/chats/{cid}`.
- `useRenameModal` sends `title` (was `topic`).

### Delete — archive the owning session
Instead of removing the chat row, the "Delete chat" action now archives the owning session via `PATCH /api/sessions/{sid} { status: "archived" }`. The api filters chats from archived sessions out of `GET /api/chats` (api#630), so the row disappears from the sidebar; archive also triggers `stopSandboxOnArchive` so we tear down running sandboxes for chats the user said they're done with. The whole thing is reversible from the admin side.

This also sidesteps the `400 Cannot delete the only chat in a session` guard that blocks chat-row deletes when a session has a single chat — archive has no such constraint.

- New `lib/sessions/archiveSession.ts` → `PATCH /api/sessions/{sid} { status: "archived" }`.
- `useDeleteChat` calls it (drops the `chatId` arg).
- `DeleteConfirmationModal` no longer passes `chatId`.

### Caveat

If a session ever has >1 chat, archiving from one row hides the others too. Today every sidebar row maps 1:1 to its own session (`POST /api/sessions` mints a session + single chat), so this is a no-op in practice — but worth knowing before we add a UI that creates additional chats inside an existing session.

### Dependencies

- Requires api#630 (filter archived sessions from `GET /api/chats`) live; until that lands, archived sessions still appear in the list.
- Base auto-rebases to `test` once chat#1756 merges.

## Test plan

- [ ] Rename a chat → sidebar shows new title after refetch. Verify via `GET /api/chats` that `title` updated server-side.
- [ ] Delete a single chat → row disappears after refetch. Verify the owning session is `status: "archived"` via `GET /api/sessions/{sid}`.
- [ ] Bulk delete multiple rows. All disappear.
- [ ] Delete a chat that's the only chat in its session (previously 400'd). Now succeeds.


arpitgupta1214:
## E2E re-test after consolidating archive change

Tested the chat preview pointed at the api#630 preview so the full archive → list-filter path is exercised end-to-end. The pointing is done client-side via `sessionStorage`:

```js
sessionStorage.setItem(
  'recoup_api_override',
  'https://api-git-feat-exclude-archived-from-chats-list-recoup.vercel.app'
);
```

**Setup**: Taylor Swift with 3 chats, all in single-chat sessions (worst-case for the old delete guard).

| Action | UI | Network call | Backend state | Sidebar |
| --- | --- | --- | --- | --- |
| Rename → "rename-take-2" | "New chat" row updates, jumps to top | `PATCH /api/sessions/ed432b84-…/chats/da78eec1-…` → 200 | `chats.title = "rename-take-2"` | 3 chats, renamed at top |
| Delete (single-chat session — previously 400'd) | Row disappears | `PATCH /api/sessions/ed432b84-…` → 200 | `sessions.status = "archived"` | 3 → 2 chats |
| Verify via api | — | `GET /api/sessions/ed432b84-…` returns `status: "archived"`; `GET /api/chats?artist_account_id=…` returns 2 chats, archived row absent | — | — |

All four PR test-plan bullets pass.

### Not exercised separately

- Bulk delete via selection mode — runs the same `useDeleteChat` hook in a loop, so per-row behavior is identical to the single-delete path above. Skipped to keep the run short.

Issue #1767, opened 2026-06-01
Post-cutover cleanup + follow-up items


sweetmantech:
Tracking issue for cleanup work that follows the chat.recoupable.com → `/api/chat/workflow` cutover ([chat#1747](https://github.com/recoupable/chat/issues/1747), shipped to prod on 2026-06-01). The cutover is fully live on `chat.recoupable.com` and `recoup-api.vercel.app`; this issue captures the remaining items that didn't need to ship with the bundle.

## Goal

Delete the legacy `/api/chat` endpoint in **api** and rename `/api/chat/workflow` → `/api/chat` so the workflow path reclaims the canonical name. Downstream: `lib/chat/handleChatStream.ts` + every other reader of `rooms`/`memories` deleted, those tables dropped, the chat client's `useChatTransport` legacy fork and `app/chat/[roomId]/page.tsx` removed, and the known UX rough edges (bootstrap spinner wait, artist-switch lifecycle, dead `/api/chats/{chatId}/artist` 404s) smoothed.

## What shipped (context)

- **api/main** has 23 commits / 13 PRs of workflow + sessions + chats + backfill + archive-filter changes (merged via [api#631](https://github.com/recoupable/api/pull/631) as `31c9d58b`)
- **chat/main** has 9 commits / 9 PRs of NewChatBootstrap + canonical URL + sidebar + artist-stamping + auto-login centralization (merged via [chat#1766](https://github.com/recoupable/chat/pull/1766) as `e8569589`)
- **prod data**: Phase 2 backfill + straggler backfill (6 additional rooms, 16 messages) all migrated cleanly to `sessions`/`chats`/`chat_messages`; 17,991 sessions with correct `artist_id`; 0 migratable stragglers remaining

## Done

- [x] **PR cleanup** — closed [chat#1754](https://github.com/recoupable/chat/pull/1754) (superseded; chat-side direct-DB approach was wrong layer) and [chat#1755](https://github.com/recoupable/chat/pull/1755) (Approach A — pivoted to the incremental cutover) with rationale comments on each.

[...]

PR #1759, opened 2026-05-30
feat(chat): stamp sessions with selected artist; refactor useArtists

arpitgupta1214:
The new-chat bootstrap now forwards the selected artist's account id as `artistId` when calling `POST /api/sessions`. The api persists it on `sessions.artist_id`, grouping the chat under the artist for the sidebar's artist filter.

- `useNewChatBootstrap` reads `selectedArtist?.account_id` from `ArtistProvider` and passes it through `createSession`.
- `CreateSessionInput` accepts an optional `artistId` field.

Pairs with api#628 (now merged) and database#27 (now merged). When no artist is selected, `artistId` is omitted and the session is created without artist linkage.

## Test plan

- [ ] Select an artist, start a new chat from `/chat`, then `GET /api/sessions/{sid}` → response has `artistId === selectedArtist.account_id`.
- [ ] No artist selected, start a new chat → response has `artistId === null`.

---
## Summary by cubic
New chats now tag sessions with the selected artist (`artistId`), and the bootstrap runs exactly once per `(artistId, orgId)` with proper loading gates to avoid duplicate POSTs. Switching or clearing the artist while viewing an existing chat hard-redirects to `/chat`; on bare `/chat`, the bootstrap re-runs in place.

- New Features
  - Send `selectedArtist.account_id` as `artistId` to `POST /api/sessions` (omit if none) so `sessions.artist_id` backs the sidebar artist filter.
  - Add optional `artistId` to `CreateSessionInput`.

- Refactors
  - Split artists logic: roster moved to `hooks/artists/useArtistsRoster` (`@tanstack/react-query` keyed on `(userId, orgId)`), selection to `hooks/artists/useArtistSelection` (override → saved → first), removed `useInitialArtists`, persisted per org.
  - Roster query is gated on Privy’s `authenticated`; `setArtists` mutates the query cache; `refetchArtists` powers `getArtists`.
  - `useArtistsRoster` surfaces react-query v5 `isPending` as `isLoading` so the disabled-query window still blocks the bootstrap, preventing early provisioning with `artistId: undefined`.
  - `useNewChatBootstrap` now delegates to `hooks/sessions/useProvisionChatSession` (`useMutation` keyed on `(artistId, orgId)` and gated on artist loading); extracted `lib/sessions/provisionChatSession` to combine `createSession` + `createSandbox`.
  - In `Artist.tsx`, detect existing-chat URLs and use `window.location.href = "/chat"` when changing or deselecting the artist; on bare `/chat`, no navigation.

[...]

PR #1769, opened 2026-06-01
fix(chat): mint one session per artist switch (#1767)

ahmednahima0-beep:
---
## Summary by cubic
Ensures exactly one chat session is minted per artist/org switch and triggers a new provision after a pending switch completes. Delays provisioning until org storage is hydrated and the artist roster is loaded.

- **Bug Fixes**
  - Added `shouldProvisionChatSession` to gate `mutate()`, dedupe same inputs, block mismatched in-flight requests, and re-provision once they settle; covered by `vitest` tests.
  - Updated `useProvisionChatSession` to react to mutation state and use `provisionInputsMatch` to ignore stale success and keep bootstrapping until inputs match.
  - Exposed `isOrgReady` in `OrganizationProvider`; `useNewChatBootstrap` now enables only when `isOrgReady` and the artist roster are loaded.

- **Refactors**
  - Moved `provisionInputsMatch` to its own module for reuse and cleaner imports; added unit tests.

<sup>Written for commit 2684de8b30e7a127ed8c543b1ac86269ef124092. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate chat session creation by ensuring provisioning skips when identical requests are pending or already succeeded.
  * Blocked new provisioning while a mismatched in-flight request exists to avoid conflicting sessions.
* **Improvements**
  * Provisioning now waits for organization data to finish initializing before starting, reducing premature or multiple setup attempts.

### DIFF
--- a/hooks/sessions/__tests__/provisionInputsMatch.test.ts
+++ b/hooks/sessions/__tests__/provisionInputsMatch.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, it } from "vitest";
+import { provisionInputsMatch } from "../provisionInputsMatch";
+
+describe("provisionInputsMatch", () => {
+  it("matches when artistId and orgId equal lastVariables", () => {
+    expect(
+      provisionInputsMatch(
+        { artistId: "artist-a", orgId: "org-1" },
+        "artist-a",
+        "org-1",
+      ),
+    ).toBe(true);
+  });
+
+  it("does not match when artistId differs", () => {
+    expect(
+      provisionInputsMatch(
+        { artistId: "artist-a", orgId: undefined },
+        "artist-b",
+        undefined,
+      ),
+    ).toBe(false);
+  });
+});
--- a/hooks/sessions/__tests__/shouldProvisionChatSession.test.ts
+++ b/hooks/sessions/__tests__/shouldProvisionChatSession.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from "vitest";
-import { shouldProvisionChatSession } from "../shouldProvisionChatSession";
+import {
+  provisionInputsMatch,
+  shouldProvisionChatSession,
+} from "../shouldProvisionChatSession";
 
 describe("shouldProvisionChatSession", () => {
   const base = {
@@ -65,5 +68,27 @@ describe("shouldProvisionChatSession", () => {
         isSuccess: true,
       }),
     ).toBe(true);
+  });
+});
+
+describe("provisionInputsMatch", () => {
+  it("matches when artistId and orgId equal lastVariables", () => {
+    expect(
+      provisionInputsMatch(
+        { artistId: "artist-a", orgId: "org-1" },
+        "artist-a",
+        "org-1",
+      ),
+    ).toBe(true);
+  });
+
+  it("does not match when artistId differs", () => {
+    expect(
+      provisionInputsMatch(
+        { artistId: "artist-a", orgId: undefined },
+        "artist-b",
+        undefined,
+      ),
+    ).toBe(false);
   });
 });
--- a/hooks/sessions/__tests__/shouldProvisionChatSession.test.ts
+++ b/hooks/sessions/__tests__/shouldProvisionChatSession.test.ts
@@ -1,8 +1,5 @@
 import { describe, expect, it } from "vitest";
-import {
-  provisionInputsMatch,
-  shouldProvisionChatSession,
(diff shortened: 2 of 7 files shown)


ahmednahima0-beep:
image

sweetmantech (approved):

sweetmantech:
## The 3×-sessions bug this PR targets is already fixed on `main` — by #1759's own merge

**TL;DR:** the headline "mints 3 sessions per artist switch" was flagged on a **#1759 preview SHA that never merged**. The version of #1759 that actually landed already closed the race. I could not reproduce 3× on production across extensive testing — it's consistently **1 session per switch**.

### Where it was fixed (specific source)

The race is the gate firing while `selectedArtist` is still null. It's closed in **`hooks/artists/useArtistsRoster.ts`** by exposing react-query's `isPending` (not `isLoading`) as the roster's loading flag — so the bootstrap gate stays closed until the roster data is actually present:

🔗 https://github.com/recoupable/chat/blob/4fad683bfd266e592c744db61a9c4bda278d2ae1/hooks/artists/useArtistsRoster.ts#L60-L86

The code even documents the exact race:
> `// Expose `isPending` (true while disabled too) as `isLoading`:`
> `// v5's `isLoading` flips false while `enabled` is false, which would`
> `// let `useNewChatBootstrap` fire with `artistId: undefined` ...`

Combined with `useArtistSelection` deriving the selection in a single `useMemo` render, `selectedArtist` is null **only while the roster is empty**, during which `enabled (authenticated && !isArtistsLoading)` is false. The `artist_id: null` orphaned session from the flag can't occur.

This shipped in **`4fad683b`** ("feat(chat): stamp sessions with selected artist; refactor useArtists (#1759)"), which is on `main`.

### Why this PR exists anyway

The 3× was flagged on the #1759 **preview** SHA `4d26b488` ([flag comment](https://github.com/recoupable/chat/pull/1759#issuecomment-4595622531)). I confirmed `4d26b488`:
- **never merged to `test`** (branch-only), and
- still used the buggy `isLoading` (`git show 4d26b488:hooks/artists/useArtistsRoster.ts` → `const { data: artists = [], isLoading } = useQuery(...)`).

The flag itself said it "could be a follow-up rather than a blocker," so #1769 was opened as that follow-up — but the **merged** #1759 (`isPending`) had already neutralized the primary race before #1769 landed.

### Live verification (current prod)

Reproduced the exact path — open chat → switch artist → `window.location.href="/chat"` hard-nav — **4×**, across: personal account, an **org with a heavy repo**, and **4× CPU + Slow 4G** throttling. Every time: **1** `POST /api/sessions` + **1** `POST /api/sandbox`. Never 3.

### Recommendation

This PR is **redundant for the headline 3× bug** (already mitigated). What it still legitimately adds is narrow defensive hardening — `shouldProvisionChatSession` blocks a new `mutate()` while a *mismatched* request is in-flight, which the current guard doesn't cover and could matter under rapid org+artist changes (well unit-tested). Suggest either:
- **close** as "already fixed by #1759's `isPending` gate" and correct the #1767 issue line, or
- **keep + re-scope** the title/description to "defensive hardening for the in-flight-mismatch edge" rather than "fixes 3× per click."

sweetmantech:
Closing as redundant for the bug it targeted. Per the #1767 analysis, the "3 sessions per artist switch" issue was already fixed on `main` by #1759's switch to react-query's `isPending` as the roster gate ([useArtistsRoster.ts L60-86](https://github.com/recoupable/chat/blob/4fad683bfd266e592c744db61a9c4bda278d2ae1/hooks/artists/useArtistsRoster.ts#L60-L86)) plus `useArtistSelection`'s single-render `useMemo` — verified on prod (1× `POST /api/sessions` + 1× `POST /api/sandbox` per switch across personal/org/throttled/hard-nav paths). Full analysis: https://github.com/recoupable/chat/pull/1769#issuecomment-4615613978

This PR remained only optional defensive hardening for the in-flight-mismatch edge; closing to keep the queue clean. Easy to revive if that edge ever surfaces.

PR #1773, opened 2026-06-03
feat(chat): typeable input while bootstrap prepares (#1767)

ahmednahima0-beep:
---
## Summary by cubic
Users can type while a new chat session provisions. Provisioning runs at the app level; typed text is restored on mount; the next session is queued after first send; and stale provisions reset once when returning.

- **New Features**
  - Added `NewChatPreparingShell` with a typeable input and “Preparing…” state; Send is disabled and announced via aria-live until ready.
  - Preserve drafted text and pass it as `initialInput` through `Chat` → `VercelChatProvider` → `useVercelChat`.
  - Replaced the full-screen spinner in `NewChatBootstrap` with the preparing shell.

- **Refactors**
  - Centralized bootstrap in `ChatSessionProvisionProvider`; `NewChatBootstrap` resets a stale provision once on mount after a prior send and clears draft input on idle/bootstrapping; `useNewChatBootstrap` is read-only.
  - `useProvisionChatSession` now returns `{ state, reset }` and uses a mint-generation counter to force fresh sessions on reset; `useVercelChat` marks a provision as consumed on first send (including query flow); updated `Providers` to include the new provider.

<sup>Written for commit c560e93d324f8ce85175ead7fb837a8ec74f4a4e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1773?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * You can type and keep a draft in the message input while a new chat session is being prepared.
  * A preparation UI now appears during session initialization showing a “Preparing…” status and disabling submission.

* **Improvements**
  * Initial input can be seeded when starting/restoring chats, improving continuity across sessions.
  * Session provisioning is more robust and can be reset/marked as consumed to avoid duplicate provisioning.

### DIFF
--- a/components/VercelChat/NewChatBootstrap.tsx
+++ b/components/VercelChat/NewChatBootstrap.tsx
@@ -1,8 +1,8 @@
 "use client";
 
 import type { UIMessage } from "ai";
-import { useState } from "react";
-import { useNewChatBootstrap } from "@/hooks/useNewChatBootstrap";
+import { useEffect, useState } from "react";
+import { useChatSessionProvision } from "@/providers/ChatSessionProvisionProvider";
 import { Chat } from "@/components/VercelChat/chat";
 import NewChatPreparingShell from "@/components/VercelChat/NewChatPreparingShell";
 
@@ -19,10 +19,27 @@ interface NewChatBootstrapProps {
 export default function NewChatBootstrap({
   initialMessages,
 }: NewChatBootstrapProps) {
-  const state = useNewChatBootstrap();
+  const { state, consumedChatId, resetProvision } = useChatSessionProvision();
   const [draftInput, setDraftInput] = useState("");
 
-  if (state.status === "ready") {
+  const isConsumedReady =
+    state.status === "ready" &&
+    consumedChatId !== null &&
+    state.chatId === consumedChatId;
+
+  useEffect(() => {
+    if (isConsumedReady) {
+      resetProvision();
+    }
+  }, [isConsumedReady, resetProvision]);
+
+  useEffect(() => {
+    if (state.status === "bootstrapping" || state.status === "idle") {
+      setDraftInput("");
+    }
+  }, [state.status]);
+
+  if (state.status === "ready" && !isConsumedReady) {
     return (
       <Chat
         id={state.chatId}
--- a/components/VercelChat/NewChatBootstrap.tsx
+++ b/components/VercelChat/NewChatBootstrap.tsx
@@ -1,31 +1,33 @@
 "use client";
 
 import type { UIMessage } from "ai";
-import { Loader } from "lucide-react";
+import { useState } from "react";
 import { useNewChatBootstrap } from "@/hooks/useNewChatBootstrap";
 import { Chat } from "@/components/VercelChat/chat";
+import NewChatPreparingShell from "@/components/VercelChat/NewChatPreparingShell";
 
 interface NewChatBootstrapProps {
   initialMessages?: UIMessage[];
 }
 
 /**
- * Thin renderer over `useNewChatBootstrap`. Mounted by
- * `app/page.tsx` (via HomePage) and `app/chat/page.tsx`; provisions
- * a recoup-api session + sandbox before mounting `<Chat>` so the
- * workflow transport (`/api/chat/workflow`) has the `sessionId` +
- * `chatId` its validator requires (recoupable/chat#1747).
+ * New-chat entry for `/` and `/chat`. Provisioning runs at app level
+ * (`ChatSessionProvisionProvider`); while ids are pending this shows a
+ * typeable input instead of a full-screen spinner. `<Chat>` mounts only
+ * when `sessionId` + `chatId` are ready (#1765).
  */
 export default function NewChatBootstrap({
   initialMessages,
 }: NewChatBootstrapProps) {
   const state = useNewChatBootstrap();
+  const [draftInput, setDraftInput] = useState("");
 
   if (state.status === "ready") {
     return (
       <Chat
         id={state.chatId}
         sessionId={state.sessionId}
+        initialInput={draftInput}
         initialMessages={initialMessages}
       />
     );
@@ -40,8 +42,6 @@ export default function NewChatBootstrap({
   }
 
   return (
-    <div className="flex size-full items-center justify-center">
-      <Loader className="size-5 animate-spin text-muted-foreground" />
-    </div>
+    <NewChatPreparingShell input={draftInput} onInputChange={setDraftInput} />
   );
 }
--- a/components/VercelChat/NewChatBootstrap.tsx
+++ b/components/VercelChat/NewChatBootstrap.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import type { UIMessage } from "ai";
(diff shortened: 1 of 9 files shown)


components/VercelChat/NewChatBootstrap.tsx
```diff
@@ -1,31 +1,56 @@
...
 
 /**
- * Thin renderer over `useNewChatBootstrap`. Mounted by
- * `app/page.tsx` (via HomePage) and `app/chat/page.tsx`; provisions
- * a recoup-api session + sandbox before mounting `<Chat>` so the
- * workflow transport (`/api/chat/workflow`) has the `sessionId` +
- * `chatId` its validator requires (recoupable/chat#1747).
+ * New-chat entry for `/` and `/chat`. Provisioning runs at app level
+ * (`ChatSessionProvisionProvider`); while ids are pending this shows a
+ * typeable input instead of a full-screen spinner. `<Chat>` mounts only
```
sweetmantech, line 16: **The “typeable input instead of a full-screen spinner” goal regresses at the shell→Chat handoff.**

`useVercelChat` still calls the message loader unchanged (`hooks/useVercelChat.ts:284-286`):
```ts
useMessageLoader(sessionId, messages.length === 0 ? id : undefined, ...)
```
When `<Chat>` mounts on `ready` with the brand-new chat id, `useMessageLoader` initializes `useState(!!chatId)` → `true`, so `ChatContentMemoized` hits `if (isLoading) return <Loader/>` and shows a **full-screen spinner** while it fetches history for a chat that has none. Net UX:

> shell (with typed text) → **full spinner (text vanishes)** → empty Chat with text restored

**Fix:** don’t load history for a bootstrap-created chat — gate the loader on the *route* chatId so it only runs on the canonical `/sessions/[sessionId]/chats/[chatId]` route:
```ts
const { chatId: routeChatId } = useParams<{ chatId?: string }>();
useMessageLoader(
  sessionId,
  messages.length === 0 && routeChatId ? id : undefined,
  userId,
  setMessages,
);
```

components/VercelChat/NewChatBootstrap.tsx
```diff
@@ -1,31 +1,56 @@
...
+  }, [state, consumedChatId, resetProvision]);
+
+  useEffect(() => {
+    if (state.status === "bootstrapping" || state.status === "idle") {
+      setDraftInput("");
+      staleResetStartedRef.current = false;
+    }
+  }, [state.status]);
 
   if (state.status === "ready") {
```
sweetmantech, line 48: **Stale consumed-session renders for a frame when returning here.**

The reset runs in a `useEffect`, which fires *after* the first paint. So when the user clicks “New Chat” after a prior send (the app-level provider still holds `ready` with `chatId === consumedChatId`), this first render returns `<Chat id={consumedChatId}>`. That mounts the already-used chat and fires a message fetch for it — briefly surfacing the previous conversation — before the reset effect swaps to a fresh session.

**Fix:** treat “ready but consumed” as not-ready *during render* so the stale Chat never mounts; keep the effect to trigger the reset:
```ts
const isConsumed =
  state.status === "ready" && consumedChatId === state.chatId;
if (state.status === "ready" && !isConsumed) {
  return <Chat id={state.chatId} sessionId={state.sessionId} initialInput={draftInput} .../>;
}
// otherwise fall through to <NewChatPreparingShell/>; the reset effect still fires
```

providers/Providers.tsx
```diff
@@ -34,11 +35,13 @@ const Providers = ({ children }: { children: React.ReactNode }) => (
               <UserProvider>
                 <OrganizationProvider>
                   <ArtistProvider>
-                    <SidebarExpansionProvider>
-                      <ConversationsProvider>
-                        <PaymentProvider>{children}</PaymentProvider>
-                      </ConversationsProvider>
-                    </SidebarExpansionProvider>
+                    <ChatSessionProvisionProvider>
```
sweetmantech, line 38: **Mounting this above the whole tree provisions a sandbox on every login for every user.**

`POST /api/sessions` + `POST /api/sandbox` now fire as soon as auth + artist roster are ready — including for users who go straight to an existing chat, files, or settings and never start a new chat. The sandbox is the expensive resource: measured `POST /api/sandbox` ≈ **10.9s cold** plus compute, so this eagerly burns a sandbox per login that often goes unused.

Worth a deliberate decision. Options:
- (a) scope provisioning to a new-chat surface rather than app-wide, or
- (b) eagerly create the **session row** but defer `createSandbox` until the new-chat view actually mounts.

If we keep it app-wide, let’s at least confirm idle sandboxes hibernate/expire cheaply.

components/VercelChat/NewChatPreparingShell.tsx
```diff
@@ -0,0 +1,75 @@
...
+          >
+            <PromptInput
+              onSubmit={(event) => event.preventDefault()}
+              className={cn(
+                "overflow-visible",
+                "rounded-2xl border border-border bg-background/70 backdrop-blur",
+                "shadow-sm",
+              )}
+            >
+              <FileMentionsInput
```
sweetmantech, line 48: **Focus seam + duplicated/reduced input.**

1. The shell’s textarea is a different DOM node from the real `ChatInput`, so focus drops at the shell→Chat swap (and a keystroke in the commit/unmount window can be lost). At minimum, autofocus on handoff.
2. This shell omits the attachments button, model select, and @-mention resolution that `ChatInput` has, and duplicates its markup. Reusing `ChatInput` in a “send-disabled” mode (e.g. an `isBootstrapPreparing` flag on the chat context) would remove the duplication **and** the focus seam at once, while keeping full input capability during preparing.

hooks/useVercelChat.ts
```diff
@@ -55,8 +58,9 @@ export function useVercelChat({
   const artistId = selectedArtist?.account_id;
   const messagesLengthRef = useRef<number>();
   const { addOptimisticConversation } = useConversationsProvider();
+  const { markProvisionConsumed } = useChatSessionProvision();
```
sweetmantech, line 61: **This couples the core chat hook to `ChatSessionProvisionProvider`.**

`useChatSessionProvision()` throws if `<Chat>` is ever rendered outside that provider (unit tests, Storybook, future embeds). Consider making the context optional — return a no-op `markProvisionConsumed` when the provider is absent — or pass `markProvisionConsumed` down as an optional prop so `useVercelChat` stays decoupled.

sweetmantech (commented):
Took a close pass focused on the new-chat preparing flow. Two correctness/UX regressions that I think block (#1 spinner flash at handoff, #2 stale consumed-session render), one infra/cost decision to make consciously (#3 eager sandbox per login), and two quality items (#4 focus seam + duplicated input, #5 hook coupling). Details inline. Overall direction is solid — keeping `sessionId` required and provisioning early is the right call; these are about the handoff edges.

PR #1872, opened 2026-07-20
feat(onboarding): state-derived onboarding router with resume-on-landing + skip-for-now

sweetmantech:
Part of recoupable/chat#1867 ("State-derived onboarding router: resume-on-landing + skip-for-now").

## What

On every authenticated landing on chat home, the app derives the account's onboarding step from the activation checkpoint predicates and routes accordingly:

- **Incomplete account** -> renders the onboarding sequence at the first unmet step, with an always-visible "Skip for now".
- **Skip** -> drops to the normal app with a dismissible checklist pinned bottom-right; "Resume setup" re-opens the sequence.
- **Fully activated account** -> normal home, never sees the sequence.

Checkpoint predicates (in sequence order), all over existing data:

| Step | Predicate | Source |
|------|-----------|--------|
| `artists` | account has >= 1 rostered artist | roster via existing `fetchArtists` (`account_artist_ids`) |
| `socials` | every rostered artist has >= 1 linked social | `account_socials` on the roster response |
| `catalog` | account has >= 1 claimed catalog | existing `getCatalogs` (`account_catalogs`) |
| `task` | account has >= 1 enabled scheduled task | existing `getTasks` (`scheduled_actions.enabled`) |

## Derived, never stored

Per the issue's architecture decision: there is **no `onboarding_step` column and no wizard cursor**. The step is a pure function of account state recomputed on every landing, so:

- drop-off/return and multi-device resume are correct for free (state lives in the DB, not the browser);
- out-of-band completion (valuation auto-claim, chat, API) auto-advances the step and can never desync;
- "skip" and "checklist dismissed" are session-scoped escape hatches (`sessionStorage`), not persisted state: the next visit re-derives and resumes by default.

The gate also fails open: while any checkpoint source is loading or errored, the normal app renders, so the sequence never flashes for activated accounts and a failed fetch never walls anyone off.

## Structure

- `lib/onboarding/getOnboardingCheckpoints.ts` - the four predicates -> checkpoint flags (pure)
- `lib/onboarding/getOnboardingStep.ts` - first unmet checkpoint -> step enum (pure, core logic)
- `lib/onboarding/getOnboardingView.ts` - {isReady, step, skipped, dismissed} -> none | sequence | checklist (pure)
- `lib/onboarding/getOnboardingStepContent.ts` - placeholder card copy + link per step
- `hooks/useOnboardingState.ts` - composes existing data hooks (`useArtistProvider`, `useCatalogs`, `useScheduledActions`) into derived state
[...]

Issue #1867, opened 2026-07-20
Onboarding sequence: state-derived resumable flow (valuation aha → roster/socials/catalog verify → first task) + activation fixes


sweetmantech:
Tracking issue for the Recoup onboarding sequence: a resumable, state-derived flow on the chat subdomain that carries every new account from the valuation aha moment through verified roster, socials, catalog, and a confirmed first scheduled task — instead of dropping them into a blank chat box. Verified problem end-to-end on prod 2026-07-20 with a fresh test signup. Sibling issues: [chat#1863](https://github.com/recoupable/chat/issues/1863) (pre-funnel catalog auto-fill + homepage honesty) and [chat#1864](https://github.com/recoupable/chat/issues/1864) (valuation pipeline resolution) — this issue owns what happens *after* a signup runs a valuation. Business context (out of scope here): all four 2026-07-13→20 signups (incl. digital@epitaph.com) stalled before their first scheduled task and none returned after day one.

> **Decision 2026-07-20 (Sweets, after the prod walkthrough + comanager competitive review — see [comment](https://github.com/recoupable/chat/issues/1867#issuecomment-5026362639)):** reframed from a set of point fixes to a **proper onboarding sequence**. Constraints agreed: (1) the sequence lives entirely on the chat subdomain — **no marketing-repo changes**; the existing "Get the full report" CTA simply redirects into it. (2) The sequence starts *after* the valuation aha, never in front of it — the 30-second zero-form dollar figure is our edge over comanager's intake-first flow and must not regress. (3) **Resume-by-default with a "skip for now" escape hatch** — incomplete accounts land on the sequence at their current step; skip drops them to the app with the checklist pinned (never a hard wall — labels/API users must be able to look around). (4) **Onboarding step is derived from account state on every landing, never stored as a wizard cursor** — the checkpoint predicates over existing tables are the source of truth, which makes drop-off/return, multi-device, and out-of-band completion (chat, API, auto-claim) correct for free. (5) For new accounts the task step is **create + confirm the first task** (pre-run its first output, then confirm the schedule) — a net-new feature establishing a recurring point of contact with every new account; auditing *existing* tasks stays out of scope (that's fleet hygiene, [chat#1846](https://github.com/recoupable/chat/issues/1846) territory).

[...]

PR #1876, opened 2026-07-20
fix: single-flight the signup account bootstrap to stop the double POST /api/accounts race (#1875)

### LINKED ISSUE #1875: Fresh web signup double-fires POST /api/accounts: races into an email-less orphan account + stale account_id (403 loops, broken first landing)
**Summary:** Every fresh web signup fires `POST /api/accounts` twice from the client. The two calls race: in the losing order, one call creates an **email-less orphan account row** and `UserProvider.userData.account_id` keeps the orphan's id while the Privy token maps to the real account — so every account-scoped fetch (`useCatalogs` → `GET /api/accounts/{id}/catalogs`) loops on `403 Access denied to specified account_id` until a full reload re-resolves the provider. Downstream, the onboarding router (#1872) correctly fails open, so the new user's **first landing never shows the onboarding sequence** — the exact moment it exists for.

**Evidence (2026-07-20, preview `chat-p8yftyrp0` = PR #1872 head + test-recoup-api; UserProvider/account-minting code untouched by that PR):**
- Signup A `sweetmantech+july2020261738@gmail.com`: two `POST /api/accounts` (both 200) → DB has real account `d2f2308d-…` (with email) **and orphan `401bdd34-…` (no email row)**; console looped `Error fetching catalogs: HTTP 403 Access denied to specified account_id` (≥5 retries); first landing showed no sequence; **reload self-healed** (provider resolved `d2f2308d`, catalogs 200, sequence rendered). Full trace in [#1872's verification comment](https://github.com/recoupable/chat/pull/1872#issuecomment-5028163186).
- Signup B `sweetmantech+july2020261819@gmail.com` (same tab, after sign-out): two `POST /api/accounts` again (reqids 647, 675) → **single account** `5604c8ee-…`, all account-scoped GETs 200, no orphan. Confirms the double-POST is deterministic and the orphan/mismatch is the race's losing branch.

**Repro:** clean browser → chat.recoupable.dev (or any preview) → sign up with a brand-new email via Privy OTP → watch the network tab: two `POST /api/accounts`; on losing the race, catalogs 403-loop + an accounts row with no `account_emails` entry.

**Expected:** one signup → exactly one account row; `userData.account_id` always matches the account the bearer token resolves to; first landing renders with working account-scoped data.

**Suspected area:** two independent client paths both ensure-account on first auth (account bootstrap on auth state change + the post-wallet-creation path — the second POST lands right after `POST auth.privy.io/v1/wallets`). Fix directions: single-flight the ensure-account call client-side, and/or make `POST /api/accounts` idempotent per Privy DID server-side so both calls return the same row.

**Done when:** a fresh signup produces exactly one `accounts` row (with email); no `403 Access denied` in console on first landing; the onboarding sequence renders on the first landing without a reload (verifiable on any preview with a new `sweetmantech+` alias). Cleanup: audit `accounts` rows with no `account_emails`/wallet linkage for orphans created by this race (`401bdd34-…` is one known instance from today).

Related: #1867 (onboarding sequence — this bug suppresses its step-1 moment), #1872 (router PR whose fail-open surfaces the symptom).

sweetmantech:
Fixes #1875. Prerequisite for #1872 per #1867's merge quality bar.

## Root cause

Both `POST /api/accounts` calls come from the **same** effect in `hooks/useUser.tsx:114-134` (pre-fix), which re-fires as Privy resolves identity in two steps. `lib/accounts/fetchOrCreateAccount.ts:20` is the only client wrapper for the endpoint, and `useUser` is its only caller, so the "two paths" from the issue trace are two firings of one effect:

1. **Post-authenticate firing:** right after `passwordless/authenticate`, `user.email.address` becomes set while `address` is still undefined (`hooks/useUser.tsx:15-16`), so the effect (deps `[email, address, getAccessToken]`) fires `POST /api/accounts` with `{ email }`.
2. **Post-wallet firing:** right after `POST auth.privy.io/api/v1/wallets` creates the embedded wallet, `user.wallet.address` populates `address`, the dep changes, and the effect fires a second `POST /api/accounts` with `{ email, wallet }` while call 1 is often still in flight.

Server-side, the two concurrent creates race read-then-write on the email linkage: both miss the lookup, both insert an `accounts` row, only one gets the `account_emails` row. The loser is the email-less orphan (`401bdd34-…` in the issue). Each `init()` closure then does last-write-wins into `setUserData`, so when the orphan-creating response lands last, `userData.account_id` holds the orphan while the bearer token resolves to the real account, and every account-scoped fetch 403-loops (`useCatalogs` → `GET /api/accounts/{id}/catalogs`).

## Before / after per path

| | Before | After |
|---|---|---|
| Post-authenticate firing | `POST {email}` immediately; could also fire before Privy resolved (no `ready` gate) | Gated on Privy `ready` so the identifying payload (email) is always known; goes through `ensureAccount` |
| Post-wallet firing | Second concurrent `POST {email, wallet}` racing the first | Serialized behind the first call's promise; the server finds the existing row and links the wallet, no second create |
| Response handling | Two racing closures both `setUserData`; stale/orphan response could win | Identical concurrent calls share one in-flight promise; distinct calls resolve in call order, and an effect-cleanup `cancelled` flag ensures only the latest run writes `userData` |

## Changes

- **`lib/accounts/ensureAccount.ts`** (new): single-flight wrapper around `fetchOrCreateAccount`. Identical concurrent calls (keyed on `email|wallet`) share one in-flight request; distinct concurrent calls are chained strictly one after another; a rejected call does not block the next.
- **`hooks/useUser.tsx`**: init effect now gates on Privy `ready`, calls `ensureAccount`, and cancels stale runs on cleanup.

## Test evidence (TDD)

`lib/accounts/__tests__/ensureAccount.test.ts` mocks the API's read-then-write semantics (concurrent creates each mint a row, exactly like the production race).

- **RED** (before `ensureAccount.ts` existed): `4 failed (4)` — the suite encodes "two concurrent bootstrap calls must produce one account creation and a consistent `account_id`".
- **GREEN** (after): `4 passed (4)`; full suite `pnpm exec vitest run`: **26 files / 87 tests passed**.

## Verification

- `pnpm exec vitest run`: 87/87 pass locally.
- `pnpm exec tsc --noEmit`: identical error set to clean `main` (verified via `git stash` baseline run; the 8 pre-existing errors are in `SpotifyDeepResearchResult.tsx` and `extractSendEmailResults.test.ts`, untouched here). No new errors.
- Prettier run on all touched files.
- **Preview verification pending:** per #1875's Done-when, sign up on this PR's preview with a fresh `sweetmantech+` alias and confirm: exactly one `POST /api/accounts` creates a row (with email), no `403 Access denied` in console on first landing, and (with #1872) the onboarding sequence renders without a reload.

## API-side follow-up (not in this PR)

Client single-flighting removes this client's race, but `POST /api/accounts` can still be double-hit by other clients (CLI, retries, multiple tabs). The server-side fix is idempotency per authenticated Privy DID (create-or-return under a uniqueness guarantee) in the `api` repo; that is the cross-repo follow-up flagged in #1875.

---
## Summary by cubic
Single-flights the signup account bootstrap to stop the double POST /api/accounts race. Prevents orphan account rows and the initial 403 loop on first landing.

- **Bug Fixes**
  - Added `lib/accounts/ensureAccount.ts`: identical concurrent calls share one in-flight request; distinct calls are serialized.
  - Updated `hooks/useUser.tsx`: gate init on Privy `ready`, use `ensureAccount`, and cancel stale effect runs so only the latest writes `userData`.
  - Added tests (`lib/accounts/__tests__/ensureAccount.test.ts`) to cover concurrent signup paths and error recovery.

<sup>Written for commit 8e1ca8095857dec91bbf5f31fb7170c79e666f86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/hooks/useUser.tsx
+++ b/hooks/useUser.tsx
@@ -6,11 +6,11 @@ import { uploadFile } from "@/lib/arweave/uploadFile";
 import { useAccount } from "wagmi";
 import { toast } from "sonner";
 import { AccountWithDetails } from "@/lib/supabase/accounts/getAccountWithDetails";
-import { fetchOrCreateAccount } from "@/lib/accounts/fetchOrCreateAccount";
+import { ensureAccount } from "@/lib/accounts/ensureAccount";
 import { updateAccountProfile } from "@/lib/accounts/updateAccountProfile";
 
 const useUser = () => {
-  const { login, user, logout, getAccessToken } = usePrivy();
+  const { login, user, logout, getAccessToken, ready } = usePrivy();
   const { address: wagmiAddress } = useAccount();
   const address = (user?.wallet?.address as Address) || wagmiAddress;
   const email = user?.email?.address;
@@ -112,13 +112,19 @@ const useUser = () => {
   };
 
   useEffect(() => {
+    // Wait for Privy to resolve so we never POST /api/accounts before the
+    // identifying payload (email) is known (chat#1875).
+    if (!ready) return;
+    if (!email && !address) return;
+    let cancelled = false;
     const init = async () => {
       const accessToken = await getAccessToken();
-      const data = await fetchOrCreateAccount({
+      const data = await ensureAccount({
         email,
         wallet: address,
         accessToken,
       });
+      if (cancelled) return;
 
       setUserData(data);
       setImage(data.image || "");
@@ -129,9 +135,11 @@ const useUser = () => {
       setRoleType(data.role_type || "");
       setCompanyName(data.company_name || "");
     };
-    if (!email && !address) return;
     init();
-  }, [email, address, getAccessToken]);
+    return () => {
+      cancelled = true;
+    };
+  }, [ready, email, address, getAccessToken]);
 
   return {
     address,
--- a/lib/accounts/__tests__/ensureAccount.test.ts
+++ b/lib/accounts/__tests__/ensureAccount.test.ts
@@ -1,0 +1,123 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import type { AccountWithDetails } from "@/lib/supabase/accounts/getAccountWithDetails";
+
+vi.mock("@/lib/accounts/fetchOrCreateAccount", () => ({
+  fetchOrCreateAccount: vi.fn(),
+}));
+
+const account = (id: string): AccountWithDetails =>
+  ({ account_id: id }) as AccountWithDetails;
+
+const tick = () => new Promise((resolve) => setTimeout(resolve, 0));
+
+/**
+ * Mimics POST /api/accounts semantics: read-then-write keyed by email.
+ * Concurrent calls both miss the read and each create a row (the chat#1875
+ * orphan bug); serialized calls resolve to the same row.
+ */
+function makeRaceyApi() {
+  const rows = new Map<string, string>();
+  let creates = 0;
+  const impl = async ({ email }: { email?: string }) => {
+    const existing = email ? rows.get(email) : undefined;
+    await tick(); // simulate network gap between read and write
+    if (existing) return account(existing);
+    creates += 1;
+    const id = `account-${creates}`;
+    if (email) rows.set(email, id);
+    return account(id);
+  };
+  return { impl, rowCount: () => rows.size, createCount: () => creates };
+}
+
+async function loadModules() {
+  vi.resetModules();
+  const { ensureAccount } = await import("@/lib/accounts/ensureAccount");
+  const { fetchOrCreateAccount } = await import(
+    "@/lib/accounts/fetchOrCreateAccount"
+  );
+  return {
+    ensureAccount,
+    fetchOrCreateAccount: vi.mocked(fetchOrCreateAccount),
+  };
+}
+
+describe("ensureAccount", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("serializes distinct concurrent calls so only one account is created", async () => {
+    const { ensureAccount, fetchOrCreateAccount } = await loadModules();
+    const api = makeRaceyApi();
+    fetchOrCreateAccount.mockImplementation(api.impl);
+
+    // Path 1: fires right after passwordless/authenticate (email only).
+    // Path 2: fires right after embedded-wallet creation (email + wallet).
+    const [first, second] = await Promise.all([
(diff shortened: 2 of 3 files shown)


sweetmantech:
**Preview verification — 2026-07-20** · preview `chat-p0f69lcfg-recoup.vercel.app` (sha-verified against PR head `8e1ca809`), Chrome DevTools, brand-new signup `sweetmantech+july2020261849@gmail.com` per #1875's Done-when.

| # | Check (from #1875 Done-when) | Expected | Actual | Result |
|---|---|---|---|---|
| 1 | Accounts created for a fresh signup | Exactly one row, with email | Supabase: single row `43872761-…` ↔ the signup email; **no orphan** (vs. pre-fix signup A which minted `401bdd34-…` with no email) | ✅ |
| 2 | `403 Access denied` in console on first landing | Zero | Zero — only a benign pre-auth 401 on `/api/sandboxes` (fires before the token exists; pre-existing, also present on main) | ✅ |
| 3 | Account-scoped fetches use the token's account | Single consistent id | All GETs (credits, subscription, agent-templates) used `43872761-…`; no cross-id requests | ✅ |
| 4 | `POST /api/accounts` behavior | No concurrent double-create | Two POSTs still occur (email-set, then wallet-set) but **serialized by `ensureAccount`** — the second resolves to the existing row instead of racing a duplicate insert. Matches the fix design; the count-to-one collapse would come from the api-side DID idempotency follow-up noted in the PR body | ✅ |

Comparison baseline: the pre-fix race evidence (two live signups, one losing the race into an orphan + 403 loop) is documented in #1875 and [#1872's verification](https://github.com/recoupable/chat/pull/1872#issuecomment-5028163186).

Verdict: #1875's Done-when criteria all pass on this preview. This PR is the merge **prerequisite for #1872** per #1867's merge quality bar — recommend reviewing/merging this first, then merging `main` into `feat/onboarding-router` for the combined first-landing test.

PR #1890, opened 2026-07-27
feat(onboarding): converge home onto /setup at the derived step (chat#1889 item 1, part 2)

sweetmantech:
Part 2 of item 1 in chat#1889 — **the keystone**. Part 1 is #1887 (repoint referrers).

## Why

Traced on `main`: **4 entry paths land on 3 parallel implementations** of the same onboarding steps.

| Entry | Lands on |
|---|---|
| Marketing valuation flow → `/catalogs/{id}` → `CatalogReportCta` | `/onboarding/first-task` |
| Valuation email | same |
| Welcome email | `/setup/*` (the interactive steps) |
| Direct to chat + **every marketing CTA** (17 bare `appUrl` refs) | home's card-based `OnboardingSequence` |

Home's cards only linked **out** to the generic app pages (`/artists`, `/artists`, `/catalogs`, `/tasks`), so a direct signup never reached the interactive steps the welcome email's `/setup/*` links do. `getOnboardingStepContent`'s own docstring admitted it: *"Placeholder card copy … until then the card links to the existing page where the step can be completed today"* — scaffolding from #1867 that was never replaced after `ConfirmRosterStep` / `VerifySocialsStep` / `FirstTaskStep` shipped.

Consequence: every setup-flow fix (Spotify typeahead, delete-social, skeleton, valuation reward) currently reaches **only** email-entry users. This PR is the multiplier on all of them.

## What changed

- **`HomePage`** forwards an incomplete account to `/setup` rather than hosting a second sequence. The skipped/pinned-checklist path is unchanged.
- **`/setup` opens at the derived step** — new `SetupEntry` + `getSetupPathForStep`. It previously hard-redirected to `/setup/artists`, so a signup who had already confirmed their roster was sent back to step 1. Renders a skeleton until every checkpoint source resolves, so a slow read can't flash the wrong step.
- **Deleted the placeholder surface**: `OnboardingSequence`, `OnboardingStepCard`, `getOnboardingStepContent` (+ its test).
- **`OnboardingCheckpointList`** reads titles from new `getOnboardingStepTitle` (the checklist was the only remaining consumer of the deleted card copy).
- **`SetupSkipLink`** carries the gate's "skip for now" escape hatch onto the `/setup` surface — without it, forwarding home would turn a deliberately soft gate into a wall. Mounted by `RosterSocialsFlow` and `FirstTaskStep`.

[...]

PR #1882, opened 2026-07-22
feat(onboarding): Spotify typeahead in verify-socials (item 3b of #1881)

sweetmantech:
## Item 3b of #1881 — Spotify typeahead in verify-socials

The verify-socials Add/Edit flow made users paste a Spotify link. This reuses the existing **`SpotifyArtistSearch`** typeahead (shipped in #1878) so users search Spotify and pick their artist instead.

### What changed
- New `components/Onboarding/SocialSearchOrPaste.tsx` — leads with the `SpotifyArtistSearch` typeahead; selecting an artist submits their `external_urls.spotify` URL, which the existing `buildSocialFixPayload` path keys as `SPOTIFY`.
- A **"Paste a link instead"** fallback stays available for the non-Spotify platforms (Instagram, TikTok, YouTube, X, Facebook, Threads, Apple Music) the Spotify search can't reach — so nothing regresses.
- Wired into both `ArtistSocialsCard` (Add a profile) and `SocialRow` (Edit) in place of the raw `SocialFixForm`.

The `onSubmit(url) => Promise<boolean>` contract is unchanged, so `useSocialFix` and the existing PATCH `profileUrls` path are untouched (DRY).

### Verification
- `pnpm exec eslint` on the changed files: clean.
- `tsc`: no new errors from these files (pre-existing test-file errors unrelated).
- `pnpm build`: compiles successfully (full page-data collection needs Supabase secrets, unavailable locally).
- `buildSocialFixPayload` tests: 6/6 pass (Spotify URL → `SPOTIFY` payload already covered).

Tracking: #1881

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Use Spotify artist typeahead in verify-socials Add/Edit so users search and pick their artist instead of pasting links. Addresses item 3b in #1881 and keeps a paste-a-link fallback for non-Spotify platforms.

- **New Features**
  - Added `components/Onboarding/SocialSearchOrPaste.tsx`, leading with `SpotifyArtistSearch` and falling back to `SocialFixForm`.
  - Integrated into `ArtistSocialsCard` and `SocialRow`, replacing direct `SocialFixForm` usage.
  - Selecting an artist submits `external_urls.spotify` to the unchanged `onSubmit(url)` contract, which maps to `SPOTIFY` via the existing path.

<sup>Written for commit abf8738c01a387ed4b774723569c7037a4f900df. Summary will update on new commits.</sup>

[...]

PR #1887, opened 2026-07-23
refactor(onboarding): retire /onboarding/* into /setup/* via config redirects (chat#1889 item 1, part 1)

sweetmantech:
## What

Today direct-chat signups render the unified `OnboardingSequence` (via `HomePage` → `useOnboardingGate`), while marketing-funnel signups deep-link to the standalone `/onboarding/first-task` mount — a second, divergent onboarding surface.

This ships the **minimal user-testable first slice** of convergence: the funnel entry now routes through the same surface.

## Change

- `app/onboarding/first-task/page.tsx` — replace the standalone `<FirstTaskStep />` mount with `redirect("/")`.

The home gate renders `OnboardingSequence` at the account's **derived** step (`getOnboardingStep` — first unmet checkpoint), so the funnel CTA always lands at the right step instead of a hard-coded `first-task`, and both entry paths converge on one experience.

## Why a redirect (not re-mounting the sequence here)

`OnboardingSequence` needs the full gate wiring (checkpoints, session skip flags, resume) that `HomePage` already owns. Redirecting to `/` reuses that single source of truth (DRY) rather than duplicating the gate on this route.

## Deferred follow-up (out of scope here)

- Fully **retire the interim standalone `/onboarding/*` mounts** (`/onboarding/first-task` once nothing links to it, plus `/onboarding/roster`).
- Repoint any external/marketing deep-links straight at `/`.

Kept as a redirect for now so any live funnel link keeps working through the cutover.

## Verification

- `pnpm exec tsc --noEmit` — clean for the changed page.
- `pnpm exec eslint app/onboarding/first-task/page.tsx` — clean.
- Behavior: visiting `/onboarding/first-task` now server-redirects to `/`, where an incomplete account resumes `OnboardingSequence` at its derived step (existing `HomePage` behavior, covered by `components/Home/__tests__/HomePage.onboarding.test.tsx`).

Targets `main`. Part of chat#1885 (converge the two onboarding surfaces).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Converges funnel and email entry on the canonical setup flow. `/onboarding/first-task` now 308-redirects to `/setup/tasks` and `/onboarding/roster` to `/setup/artists`; updates the catalog report CTA to `/setup/tasks`. Aligns with `chat#1889` and progresses `chat#1885`.

- **Refactors**
  - Move redirects to 308 edge rules in `next.config.mjs` and delete the `/onboarding/*` page stubs.
[...]

PR #1886, opened 2026-07-23
feat(onboarding): render social avatar on verify-socials SocialRow

sweetmantech:
## What

In the verify-socials onboarding step, `SocialRow` rendered platform + `@handle · N followers` but no profile image, even though the `avatar` field is now populated. This renders each connected social's avatar so users can **visually confirm the auto-match**.

## Changes

- `components/Onboarding/SocialRow.tsx` — add the shared shadcn `Avatar` (circular, `radius-full` per DESIGN.md) before the platform/handle text.
  - `AvatarImage` from `social.avatar` when present.
  - `AvatarFallback` to the handle/platform initial when `avatar` is null — achromatic chrome, no color introduced.

Render-side only; mirrors the sibling `RosterArtistRow` avatar pattern.

## Verification

- `pnpm exec tsc --noEmit` — clean for `SocialRow.tsx`.
- `pnpm exec eslint components/Onboarding/SocialRow.tsx` — clean.
- Screenshot: **pending** — the verify-socials step requires an authenticated account with a claimed roster + matched socials to reach; not captured in this render-only change. Follow-up visual check recommended on preview.

Targets `main`. Part of chat#1885 (converge/polish onboarding backlog — verify-socials pfp row).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Render profile images for connected socials in the verify-socials step so users can visually confirm the auto-match. Adds the shared shadcn `Avatar` to `components/Onboarding/SocialRow.tsx`, showing `social.avatar` before the text and falling back to the uppercase first letter of the handle/platform when missing (refs chat#1885).

<sup>Written for commit 64c7474c01e8f86fc886363c5ef803aa865546e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1886?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1892, opened 2026-07-27
feat(onboarding): add artists by Spotify typeahead, not free text (chat#1889 row 8)

sweetmantech:
Matrix row 8 in chat#1889 — the highest-value item after the convergence chain. Independent of #1890/#1891 (different files), so it can be reviewed in parallel.

## Why: the payoff was structurally unreachable for most signups

Live cohort from `email_send_log`, 2026-07-23 → 2026-07-26:

| | |
|---|---|
| Welcome emails sent | **9**, to 9 distinct accounts |
| With a roster + catalog (valuation-funnel signups) | **3** |
| With **zero artists, zero catalogs, no valuation** | **6** |

The welcome email fires on account creation regardless of whether a valuation preceded it, so two-thirds of the cohort are cold-start signups. Their only forward path is `AddArtistForm` — which took a **free-text name** and posted only `{ name }`. No Spotify id → no catalog → no valuation. The number those users were emailed about could never exist.

## What changed

- **`AddArtistForm`** uses `SpotifyArtistSearch` — the same typeahead verify-socials adopted in #1882 — so a pick resolves to a real Spotify id, profile URL, and avatar.
- **`useAddRosterArtist`** now takes `{ profileUrl, image }` and attaches them via the existing `PATCH /api/artists/{id}` path after creation, because `POST /api/artists` accepts only `name` / `account_id` / `organization_id`. Reuses `buildSocialFixPayload` for the platform detection rather than hand-rolling the shape. Enrichment failure is deliberately non-fatal — the artist is already on the roster and socials can still be fixed in the next step.
- **`ConfirmRosterStep` copy**: an empty roster was told *"We set these artists up from your valuation"* — false for 6 of 9 signups. It now reads as an instruction to search, and the empty state names the reward.

No api change needed, so no docs/contract change either — this rides two existing endpoints.

## Verification

TDD, red → green:

\`\`\`
# RED — free-text field still present, no typeahead
Test Files  1 failed (1)
     Tests  2 failed (2)

# GREEN
pnpm exec vitest run components/Onboarding lib/onboarding
  Test Files  18 passed (18)
       Tests  70 passed (70)
\`\`\`

The new test asserts the free-text `Artist name` field is **gone**, the Spotify searchbox is present, and picking a result calls `addArtist` with the resolved profile URL + avatar.

`pnpm exec tsc --noEmit` clean for the touched files. Preview click-through pending (needs an authed empty-roster account).

Tracked in chat#1889 (matrix row 8).

---
## Summary by cubic
Replace free-text artist input with a Spotify typeahead in onboarding so added artists have a real Spotify profile and avatar. This unblocks cold-start signups so they can reach catalog and valuation (chat#1889, row 8).

- **New Features**
  - `AddArtistForm` now uses `SpotifyArtistSearch`; selecting a result calls `addArtist(name, { profileUrl, image })`.
  - `useAddRosterArtist` attaches the Spotify profile and image via `PATCH /api/artists/{id}` using `buildSocialFixPayload`, then refreshes the roster.
  - `ConfirmRosterStep` copy prompts Spotify search when empty and only credits a valuation if the account actually has a catalog.

- **Bug Fixes**
  - Enrichment failures are non-fatal: `saveArtist` errors are caught so the form closes and the roster refreshes, avoiding duplicates.
  - Removed em dashes from roster and socials copy (including button text) to improve tone; enforced via `getConfirmRosterCopy` tests.

<sup>Written for commit 617d7f930dc3da4ec5b0619c2c523b8260db95de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added Spotify artist search for selecting artists during onboarding.
  * When available, automatically attaches an artist’s Spotify profile link and artwork.
  * Updated the “Confirm your roster” message to be tailored to the number of artists and whether valuation is available.
* **Bug Fixes**
  * Artist additions now retain available Spotify profile and artwork details, and the flow closes automatically after a successful add.

### DIFF
--- a/components/Onboarding/AddArtistForm.tsx
+++ b/components/Onboarding/AddArtistForm.tsx
@@ -1,24 +1,31 @@
 "use client";
 
 import { useState } from "react";
-import { Loader, Plus } from "lucide-react";
+import { Plus } from "lucide-react";
 import { Button } from "@/components/ui/button";
-import { Input } from "@/components/ui/input";
+import SpotifyArtistSearch from "@/components/Artists/SpotifyArtistSearch";
 import { useAddRosterArtist } from "@/hooks/onboarding/useAddRosterArtist";
+import type { SpotifyArtistSearchResult } from "@/types/spotify";
 
-/** Inline "add another artist" form for multi-artist managers. */
+/**
+ * "Add an artist" for the onboarding roster step, by Spotify typeahead.
+ *
+ * It used to take a free-text name, which produced an artist with no Spotify id
+ * — so no catalog and therefore no valuation could follow, and the payoff was
+ * structurally unreachable for anyone who arrived without a funnel valuation
+ * (chat#1889). Picking a real Spotify artist resolves the id, profile URL, and
+ * avatar in one step, reusing the same typeahead as verify-socials (chat#1882).
+ */
 const AddArtistForm = () => {
   const { addArtist, isAdding } = useAddRosterArtist();
   const [isOpen, setIsOpen] = useState(false);
-  const [name, setName] = useState("");
 
-  const handleSubmit = async (e: React.FormEvent) => {
-    e.preventDefault();
-    const added = await addArtist(name);
-    if (added) {
-      setName("");
-      setIsOpen(false);
-    }
+  const handleSelect = async (artist: SpotifyArtistSearchResult) => {
+    const added = await addArtist(artist.name, {
+      profileUrl: artist.external_urls.spotify,
+      image: artist.images?.[0]?.url,
+    });
+    if (added) setIsOpen(false);
   };
 
   if (!isOpen) {
@@ -36,27 +43,23 @@ const AddArtistForm = () => {
   }
 
   return (
-    <form onSubmit={handleSubmit} className="flex items-center gap-2">
-      <Input
+    <div className="flex flex-col gap-2">
+      <SpotifyArtistSearch
         autoFocus
-        value={name}
-        onChange={(e) => setName(e.target.value)}
-        placeholder="Artist name"
-        disabled={isAdding}
-        aria-label="Artist name"
+        isBusy={isAdding}
+        onSelect={(artist) => void handleSelect(artist)}
       />
-      <Button type="submit" disabled={isAdding || !name.trim()}>
-        {isAdding ? <Loader className="size-4 animate-spin" /> : "Add"}
-      </Button>
       <Button
         type="button"
         variant="ghost"
+        size="sm"
+        className="self-start text-muted-foreground"
         disabled={isAdding}
         onClick={() => setIsOpen(false)}
       >
         Cancel
       </Button>
-    </form>
+    </div>
   );
 };
 
--- a/components/Onboarding/ConfirmRosterStep.tsx
+++ b/components/Onboarding/ConfirmRosterStep.tsx
@@ -58,7 +58,7 @@ const ConfirmRosterStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
         disabled={isLoading || artists.length === 0}
         onClick={onConfirmed}
       >
-        {artists.length > 1 ? "These are my artists" : "This is my artist"} —
+        {artists.length > 1 ? "These are my artists" : "This is my artist"},
         continue
       </Button>
     </section>
--- a/components/Onboarding/ConfirmRosterStep.tsx
+++ b/components/Onboarding/ConfirmRosterStep.tsx
@@ -3,6 +3,8 @@
 import { Button } from "@/components/ui/button";
 import { Skeleton } from "@/components/ui/skeleton";
 import { useArtistProvider } from "@/providers/ArtistProvider";
+import useCatalogs from "@/hooks/useCatalogs";
+import { getConfirmRosterCopy } from "@/lib/onboarding/getConfirmRosterCopy";
 import AddArtistForm from "./AddArtistForm";
 import RosterArtistRow from "./RosterArtistRow";
 
@@ -13,7 +15,9 @@ import RosterArtistRow from "./RosterArtistRow";
  */
 const ConfirmRosterStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
   const { sorted, isLoading } = useArtistProvider();
+  const { data: catalogsData } = useCatalogs();
   const artists = sorted.filter((artist) => !artist.isWorkspace);
+  const hasValuation = (catalogsData?.catalogs?.length ?? 0) > 0;
 
   return (
     <section className="flex flex-col gap-6">
@@ -21,13 +25,11 @@ const ConfirmRosterStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
         <h1 className="text-2xl font-semibold text-foreground">
           Confirm your roster
         </h1>
-        {/* Two-thirds of welcome-email signups arrive with an empty roster and
-            never ran a valuation, so claiming we set one up from "your
-            valuation" was simply false for them (chat#1889). */}
         <p className="text-sm text-muted-foreground mt-1">
-          {artists.length === 0
-            ? "Search Spotify for the artists you manage. Recoup uses them to measure your catalog, so pick the real profile."
-            : `We set ${artists.length === 1 ? "this artist" : "these artists"} up from your valuation. Add anyone else you manage — you can verify their socials next.`}
+          {getConfirmRosterCopy({
+            artistCount: artists.length,
+            hasValuation,
+          })}
         </p>
       </div>
 
--- a/components/Onboarding/ConfirmRosterStep.tsx
+++ b/components/Onboarding/ConfirmRosterStep.tsx
@@ -21,10 +21,13 @@ const ConfirmRosterStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
         <h1 className="text-2xl font-semibold text-foreground">
           Confirm your roster
         </h1>
+        {/* Two-thirds of welcome-email signups arrive with an empty roster and
+            never ran a valuation, so claiming we set one up from "your
+            valuation" was simply false for them (chat#1889). */}
         <p className="text-sm text-muted-foreground mt-1">
-          We set {artists.length === 1 ? "this artist" : "these artists"} up
-          from your valuation. Add anyone else you manage — you can verify their
-          socials next.
+          {artists.length === 0
+            ? "Search Spotify for the artists you manage. Recoup uses them to measure your catalog, so pick the real profile."
+            : `We set ${artists.length === 1 ? "this artist" : "these artists"} up from your valuation. Add anyone else you manage — you can verify their socials next.`}
         </p>
       </div>
 
@@ -36,7 +39,8 @@ const ConfirmRosterStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
           </>
         ) : artists.length === 0 ? (
           <p className="text-sm text-muted-foreground p-4 rounded-xl border border-border">
-            No artists on your roster yet. Add your first artist below.
+            No artists on your roster yet. Add your first artist below to get
+            your catalog valued.
           </p>
         ) : (
           artists.map((artist) => (
--- a/components/Onboarding/VerifySocialsStep.tsx
+++ b/components/Onboarding/VerifySocialsStep.tsx
@@ -24,7 +24,7 @@ const VerifySocialsStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
         </h1>
         <p className="text-sm text-muted-foreground mt-1">
           These are the profiles we matched. Fix any that point at the wrong
-          account — reports and tasks pull from them.
+          account. Reports and tasks pull from them.
         </p>
       </div>
 
@@ -40,7 +40,7 @@ const VerifySocialsStep = ({ onConfirmed }: { onConfirmed: () => void }) => {
       </div>
 
       <Button type="button" className="w-full" onClick={onConfirmed}>
-        Looks good — continue
+        Looks good, continue
       </Button>
     </section>
   );
--- a/components/Onboarding/__tests__/AddArtistForm.test.tsx
+++ b/components/Onboarding/__tests__/AddArtistForm.test.tsx
@@ -1,0 +1,58 @@
+// @vitest-environment jsdom
+import React from "react";
+import { fireEvent, render, screen } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import AddArtistForm from "@/components/Onboarding/AddArtistForm";
+
+const addArtist = vi.fn();
+
+const SPOTIFY_RESULT = {
+  id: "3TVXtAsR1Inumwj472S9r4",
+  name: "Drake",
+  external_urls: { spotify: "https://open.spotify.com/artist/3TVXtAsR1Inumwj472S9r4" },
+  images: [{ url: "https://i.scdn.co/image/drake.jpg" }],
+  followers: { total: 92000000 },
+};
+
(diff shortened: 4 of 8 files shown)


sweetmantech:
## Preview verification — rebased onto `main`, verified end-to-end

Rebased off `main` (the branch was 3 behind: #1887, #1886, #1890) and force-pushed as `dae37413`. Zero file overlap with those three, so the rebase was clean and the diff is unchanged at **+129/−31 across 4 files**.

- **Preview:** `https://chat-mh0bwz0gc-recoup.vercel.app` (deployment `5640583931`, built from `dae37413`)
- **Tests:** `pnpm exec vitest run components/Onboarding lib/onboarding` → **73 passed / 19 files** (was 70/18 pre-rebase; `main` contributed `getOnboardingStepTitle` + `getSetupPathForStep`)
- **Types:** `pnpm exec tsc --noEmit` → **13 errors on the branch, 13 on `main`** — measured on both, so this PR adds none
- **Account:** fresh signup `sweetman+july2820260924@recoupable.com`, account `57f79840-2c71-4da3-bd00-f916df6054eb` — a genuine zero-artist, zero-catalog, never-valued cold start, i.e. the exact 6-of-9 cohort this PR targets

### Results

| # | Check | Expected | Actual | |
|---|-------|----------|--------|---|
| 1 | Free-text `Artist name` input | gone | no such node in the a11y tree | ✅ |
| 2 | Replacement control | Spotify typeahead | `textbox "Search Spotify for an artist"` + `listbox "Spotify artist results"` | ✅ |
| 3 | `Add` submit button | gone, `Cancel` remains | exactly that | ✅ |
| 4 | Empty-roster copy | new string | "Search Spotify for the artists you manage… pick the real profile." | ✅ |
| 5 | Empty-state copy | new string | "No artists on your roster yet. Add your first artist below to get your catalog valued." | ✅ |
| 6 | Typeahead results | real Spotify artists | 20 results w/ avatars + follower counts via `GET /api/spotify/search` | ✅ |
| 7 | Fields the PR reads | populated | `external_urls.spotify` = `…/artist/43qxAkuKFB6fMNSeS5dO7Z`, `images[0].url` = `i.scdn.co/…f04af436`, `followers.total` = 1,797,355 | ✅ |
| 8 | Create call | `POST /api/artists` | **201** | ✅ |
| 9 | Enrichment call | `PATCH /api/artists/{id}` | **200** | ✅ |
| 10 | PATCH payload | profileUrls + image | `{"profileUrls":{"SPOTIFY":"https://open.spotify.com/artist/43qxAkuKFB6fMNSeS5dO7Z"},"image":"https://i.scdn.co/image/ab6761610000e5ebf04af436ecc9ce9b74d715eb"}` | ✅ |
| 11 | **Spotify id persisted** | yes | `account_socials[0].profile_url` = `open.spotify.com/artist/43qxAkuKFB6fMNSeS5dO7Z`, `type: SPOTIFY` | ✅ |
| 12 | Roster row avatar | Spotify image | renders; `artist.image` set to the scdn URL | ✅ |
| 13 | Form collapses on success | `setIsOpen(false)` | collapsed, roster refreshed to 1 artist | ✅ |

**The headline acceptance criterion — "an artist added during setup carries a Spotify id" — is met.** Verified at the data layer, not just visually: `GET /api/artists` returns the id `43qxAkuKFB6fMNSeS5dO7Z` on the created artist.

Deliberate negative probes against the enrichment endpoint:

| Probe | Result |
|---|---|
| `PATCH /api/artists/{id}` no auth | **401** `Exactly one of x-api-key or Authorization must be provided` |
| `PATCH /api/artists/not-a-uuid` | **400** `id must be a valid UUID` |
| `PATCH /api/artists/{unowned-uuid}` | **403** `Forbidden` |

### Screenshots

**Cold-start empty roster (authenticated) — both new strings**
authed empty roster

**Typeahead with live Spotify results**
typeahead results

**After selection — avatar + "1 social profile matched"**
artist added

**Verify-socials step**
verify socials

---

## 🔴 One defect, in scope — enrichment failure is not actually non-fatal

The docstring on `useAddRosterArtist` claims:

> Enrichment failing is non-fatal — the artist is already on the roster and its socials can still be fixed in the next step.

The code doesn't do that. `await saveArtist(...)` sits **inside the same `try`** as `createRosterArtist`, so a PATCH failure lands in the shared `catch`:

```ts
const artist = await createRosterArtist(accessToken, trimmed, selectedOrgId);
if (artist?.account_id && (payload || options.image)) {
  await saveArtist(accessToken, artist.account_id, {...});   // throws ⇒ shared catch
}
await getArtists();      // ← never runs
return true;             // ← never runs
```

On a PATCH failure the user gets `toast.error(...)`, `addArtist` returns `false`, so `setIsOpen(false)` never fires and `getArtists()` never refreshes — **but `POST /api/artists` already succeeded.** Net effect: the artist exists server-side, the UI says "Failed to add artist" and keeps the form open, and the obvious user response (pick again) **creates a duplicate**.

This isn't hypothetical — the probes above show the PATCH has three live failure modes (401 on a stale token, 400, 403), plus ordinary network flake. The 403 path is worth weighing given the connector-authority hardening in #1860 / #1865.

Fix is small and keeps the docstring honest:

```ts
if (artist?.account_id && (payload || options.image)) {
  try {
    await saveArtist(accessToken, artist.account_id, {
      ...(payload ? { profileUrls: payload.profileUrls } : {}),
      ...(options.image ? { image: options.image } : {}),
    });
  } catch {
    // Non-fatal by design: the artist is created and its socials are
    // fixable in verify-socials. Swallow so the roster still refreshes.
  }
}
```

Worth a test pinning it — "PATCH rejects ⇒ `addArtist` still resolves `true` and the roster refreshes" — since the current suite doesn't cover the partial-failure path.

## 🟡 The copy fix only covers the zero-artist branch

`ConfirmRosterStep` was changed for `artists.length === 0`. But as soon as the cold-start user adds their first artist **through this PR's own typeahead**, the ternary falls through to the original string and they're told:

> "We set this artist up from your valuation. Add anyone else you manage — you can verify their socials next."

on an account that has never run a valuation. (Visible in the third screenshot above — that's the same fresh account, one artist in.) The PR body lists "fixes the false 'from your valuation' copy" as a goal, and it's half-fixed: false at 0 artists is gone, false at ≥1 remains. A `hasValuation` / catalog-count condition rather than a roster-length one would settle it.

## ⚪ Pre-existing, not a regression — `@artist · 0 followers` (#1897)

Verify-socials renders `Spotify @artist · 0 followers` for an artist with 1.8M followers, with a fallback initial instead of the social avatar. The PATCH sends a correct absolute profile URL; the api derives `username` from a URL path segment and stores `followerCount: 0`, `avatar: null`.

This is the gap #1886 flagged and #1897 owns — **this is the fourth independent sighting, and the first confirmed through the new typeahead path**, which reinforces the note on #1886 that #1897's scope must cover the **handle**, not just pfp + follower count. Note the artist-level `image` is set correctly by this PR; it's only the social row that's bare.

## ⚪ Scope note — the gate is removed, the payoff is not yet reached

Completing the flow leaves the account at **"Roster verified"** with `GET /api/accounts/{id}/catalogs` → `{"catalogs": []}` and `/catalogs` rendering "No catalogs found." So the chain is now *unblocked* (the Spotify id exists, which was the hard blocker) but nothing in the setup flow creates the catalog or shows a valuation — the flow is still "Step 1 of 2" and ends on the generic green check.

That's chat#1895 (row 7, valuation reward) and row 8 (persist `est_catalog_value`), not this PR. Recording it so row 6 isn't mistaken for delivering the payoff on its own.

no catalogs

---

### Not verified

- The enrichment-failure path was probed at the endpoint level but **not forced through the UI**, so the 🔴 above is read from the code plus live failure modes rather than observed in-browser.
- No bot review exists on this PR: CodeRabbit hit the org review limit ("next review available in 37 minutes") at 04:48 on 07-27 and never re-ran; Bugbot is disabled on the account.

### Recommendation

Core change works and is worth landing. I'd fix the 🔴 before merge (few lines + one test) and decide on the 🟡 — either widen the condition here or split it out. Verified on `dae37413`.

sweetmantech:
## Both findings fixed — `fa7a630e`

Pushed on top of the rebase. TDD both, RED confirmed before any implementation.

### 🔴 Enrichment is now genuinely non-fatal

`saveArtist` moved into its own `try`/`catch` inside the existing block, so a failing `PATCH /api/artists/{id}` no longer takes the shared catch.

```ts
if (artist?.account_id && (payload || options.image)) {
  try {
    await saveArtist(accessToken, artist.account_id, { ... });
  } catch {
    // Swallowed on purpose. POST /api/artists already succeeded, so
    // surfacing this would leave the form open over an artist that
    // exists — and picking again would create a duplicate. The socials
    // are still fixable in the next step.
  }
}
```

New `hooks/onboarding/__tests__/useAddRosterArtist.test.tsx` (the hook had no test file at all) pins three behaviours:

| Test | Asserts |
|---|---|
| enriches a created artist | `saveArtist` called with `{profileUrls: {SPOTIFY: …}, image: …}`, roster refreshed |
| **treats a failed enrichment as non-fatal** | `saveArtist` rejects ⇒ `addArtist` still resolves `true`, `getArtists()` still runs, **no error toast** |
| still reports failure when the artist itself can't be created | `createRosterArtist` rejects ⇒ resolves `false`, toast fires, `getArtists()` **not** called |

The middle one is the regression guard. RED output before the fix, which is the defect reproduced:

```
FAIL  hooks/onboarding/__tests__/useAddRosterArtist.test.tsx > treats a failed enrichment as non-fatal
AssertionError: expected false to be true
  - Expected: true
  + Received: false
```

The third test matters as much as the second — it proves the new `catch` didn't swallow real create failures along with enrichment ones.

### 🟡 The valuation credit is now keyed on a valuation, not on roster length

Copy logic extracted to a pure `lib/onboarding/getConfirmRosterCopy.ts` and keyed on `hasValuation` (derived from `useCatalogs()` — a catalog is the artifact a valuation leaves behind) rather than `artists.length`:

| Roster | Valuation ran | Copy |
|---|---|---|
| 0 artists | — | "Search Spotify for the artists you manage…" |
| ≥1 | **yes** | "We set {this artist/these artists} up from your valuation. Add anyone else you manage — you can verify their socials next." |
| ≥1 | **no** | "Add anyone else you manage — you can verify their socials next." |

Pulling it out of the JSX means the three branches are unit-testable without rendering, matching how `getOnboardingStepTitle` / `getSetupPathForStep` already live in `lib/onboarding/`. The test asserts the negative directly — for rosters of 1, 2 and 5 with no valuation, the string must not match `/valuation/i`.

### Verified on preview `chat-hpp1arx60-recoup.vercel.app` (built from `fa7a630e`)

Re-checked on the **same account from the first pass** — `sweetman+july2820260924@recoupable.com`, now sitting at 1 artist / 0 catalogs, which is precisely the state that produced the false claim. Same URL, same account, before and after:

| | Copy rendered |
|---|---|
| `dae37413` (before) | "We set this artist up from your valuation. Add anyone else you manage — you can verify their socials next." |
| `fa7a630e` (after) | "Add anyone else you manage — you can verify their socials next." |

copy fixed

### Suite

```
pnpm exec vitest run components/Onboarding lib/onboarding hooks/onboarding
  Test Files  21 passed (21)
       Tests  79 passed (79)
```

Up from 73/19 at the start of this pass. `pnpm exec tsc --noEmit` → **13 errors, identical to `main`'s 13**; `pnpm exec eslint` clean on all five touched files.

### Still not verified

- The `hasValuation: true` branch is unit-tested but **not** browser-confirmed — this account has no catalog, and creating one to prove a copy string wasn't worth the side effects. The funnel-signup path renders it today on `main`, unchanged by this PR.
- The `@artist · 0 followers` gap is untouched here and, per the audit below, has **no PR fixing it** — chat#1897 explicitly scopes it out ("needs an api resolve path that does not exist yet"), and there is no api PR. It's being given its own row in chat#1889 rather than staying a parenthetical inside row 17.

sweetmantech:
## Copy pass + in-browser verification of both fixes — `617d7f93`

Em dashes read as machine-written and cost trust, so they're out of the setup copy. Verified on preview `chat-b94uye04u-recoup.vercel.app`, built from `617d7f93`, on the same cold-start account as the previous passes (`sweetman+july2820260924@recoupable.com`).

### Copy changes

| Where | Before | After |
|---|---|---|
| `getConfirmRosterCopy` tail | "Add anyone else you manage **—** you can verify their socials next." | "Add anyone else you manage**,** you can verify their socials next." |
| `ConfirmRosterStep` CTA | "This is my artist **—** continue" | "This is my artist**,** continue" |
| `VerifySocialsStep` sub-head | "…point at the wrong account **—** reports and tasks pull from them." | "…point at the wrong account**.** Reports and tasks pull from them." |
| `VerifySocialsStep` CTA | "Looks good **—** continue" | "Looks good**,** continue" |

A unit test now pins it, so it can't creep back:

```ts
it("uses no em or en dashes in any branch", () => {
  for (const hasValuation of [true, false])
    for (const artistCount of [0, 1, 2, 5])
      expect(getConfirmRosterCopy({ artistCount, hasValuation })).not.toMatch(/[—–]/);
});
```

Asserted live rather than by eye — scanning the rendered `innerText` of both steps for `/[—–]/` returns **zero matches** on each.

**Roster step** — the copy line, and the singular CTA:

roster copy

**Socials step** — sub-head and CTA:

socials copy

### 🔴 The non-fatal enrichment fix, now verified in-browser

The previous comment listed this as read-from-code rather than observed, since the failure couldn't be triggered through the UI. Closing that gap: `window.fetch` was patched in the page to force **exactly the 403 the live endpoint returns for an unowned id** on any `PATCH /api/artists/*`, then an artist was added through the typeahead.

Forced failure confirmed intercepted:

```
PATCH https://test-recoup-api.vercel.app/api/artists/c2425574-7306-4f4b-b1b9-cce8ea20d6a9  →  403 Forbidden (injected)
```

| Assertion | Result |
|---|---|
| Artist appears on the roster (`getArtists()` ran) | ✅ "Selena" rendered |
| Add form collapsed (`addArtist` resolved `true`) | ✅ `formStillOpen: false` |
| No error toast | ✅ no "Failed to add artist" anywhere in the DOM |
| Degrades honestly | ✅ row reads "No socials matched yet" — fixable in the next step, exactly as the docstring promises |
| Plural CTA branch | ✅ "These are my artists, continue" |

On `dae37413` this same sequence produced "Failed to add artist", left the form open, and left Selena off the roster despite `POST /api/artists` having already created her — the duplicate trap. It now degrades to an un-enriched roster row, which is the documented contract.

non-fatal enrichment

### Suite

```
pnpm exec vitest run components/Onboarding lib/onboarding hooks/onboarding
  Test Files  21 passed (21)
       Tests  80 passed (80)
```

`pnpm exec tsc --noEmit` → **13, unchanged from `main`**. `eslint` clean on all touched files.

### Deliberately left alone

`FirstTaskStep.tsx` L45 and `buildFirstTaskParams.ts` L26 (`"Weekly Catalog Report — {artist}"`) carry the same em dash. Not fixed here because **[chat#1898](https://github.com/recoupable/chat/pull/1898) is editing `FirstTaskStep.tsx`** and this would conflict with it. Confirmed safe to change on its own terms — `findExistingWeeklyReportTask` matches on `task.enabled`, not on the title — so it's a clean pickup for #1898 or a follow-up. Everything else matching `—` under `components/Onboarding/` is code comments, not product text.

PR #1895, opened 2026-07-27
feat(onboarding): end setup on the catalog valuation (chat#1889 row 10)

sweetmantech:
Matrix row 10 in chat#1889 — **the payoff the issue is named for**. Independent of the convergence chain.

## Why

Two dead ends on the same number:

1. Finishing roster + socials ended on a generic green `CheckCircle2` and a "Back to Recoup" button.
2. `/setup/valuation` was a bare `redirect("/catalogs")` — so the welcome email's **"See your baseline valuation"** link had no real destination.

A signup converts on a valuation and then never sees it again inside the product.

## What changed

- **`RosterVerifiedPanel`** renders the account's valuation using the **existing** homepage hero (`useHomeValuation` + `ValuationHero`) — no second valuation surface — then points at the weekly report that keeps it measured.
- **New `SetupValuation`** mounts at `/setup/valuation` with the same hero, and falls back to `/catalogs` only when the account genuinely has no catalog to value.
- The **no-valuation fallback** now points at the step that unlocks one (`/setup/catalog`) instead of dead-ending on "Back to Recoup".

## Verification

TDD, red → green:

\`\`\`
# RED — panel still renders the generic check, no valuation
× ends setup on the catalog valuation, not a generic green check
  Tests  1 failed | 1 passed (2)

# GREEN
pnpm exec vitest run components/Onboarding
  Test Files  1 passed (1)
       Tests  2 passed (2)
\`\`\`

Tests assert both directions: the valuation figure and measured-track count render when a valuation exists (`$1.4M`, `42 tracks measured`), and the confirmation state still appears when it does not — so a cold-start account can't hit an empty hero.

`pnpm exec tsc --noEmit` clean for the touched files. Preview click-through pending (needs an authed account with a measured catalog).

Tracked in chat#1889 (matrix row 10).

---
## Summary by cubic
End onboarding on the catalog valuation and make `/setup/valuation` show the same valuation instead of redirecting. Delivers chat#1889 row 10 and adds a guard to prevent cold-load bounces.

- **New Features**
  - `RosterVerifiedPanel` now shows the account’s valuation using `ValuationHero` + `useHomeValuation`, with a CTA to `/setup/tasks`. If no valuation, show confirmation with a CTA to `/setup/catalog`.
  - New `SetupValuation` at `/setup/valuation` renders the same hero and weekly report CTA; only redirects to `/catalogs` when the account has no catalog. Tests cover the hero view, fallbacks, and `/setup/valuation` redirects.

- **Bug Fixes**
  - Gate `/setup/valuation` on `isPending` (not `isLoading`) from `useCatalogs` to avoid redirecting before auth enables the query, preventing cold-load bounces from the welcome email link.

<sup>Written for commit 61911a6ad459a1e162f5473311878faddfd9aad1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1895?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a baseline valuation step to the onboarding flow.
  * Displays valuation details, artist information, and measured track counts.
  * Added a link to set up weekly reports after valuation is available.
  * Added loading placeholders while valuation and catalog information loads.
  * Routes users without an existing catalog to catalog setup.

* **Bug Fixes**
  * Welcome email links now open the intended valuation onboarding experience instead of redirecting immediately.

### DIFF
--- a/app/setup/valuation/page.tsx
+++ b/app/setup/valuation/page.tsx
@@ -1,11 +1,12 @@
-import { redirect } from "next/navigation";
+import SetupValuation from "@/components/Onboarding/SetupValuation";
 
 /**
- * `/setup/valuation` — welcome email step 4 ("See your baseline valuation").
- * The baseline valuation is the catalog report (`/catalogs/[catalogId]`), which
- * is not addressable without a catalog id here, so route to the catalog list.
- * Revisit if a dedicated valuation landing is built.
+ * `/setup/valuation` — the welcome email's payoff link. Renders the account's
+ * baseline valuation (chat#1889) instead of the bare `/catalogs` redirect it
+ * used to be, which left the email's "See your baseline valuation" link with no
+ * real destination. Reading the valuation needs client state, so the page mounts
+ * `SetupValuation`.
  */
 export default function SetupValuationPage() {
-  redirect("/catalogs");
+  return <SetupValuation />;
 }
--- a/components/Onboarding/RosterVerifiedPanel.tsx
+++ b/components/Onboarding/RosterVerifiedPanel.tsx
@@ -3,25 +3,66 @@
 import Link from "next/link";
 import { CheckCircle2 } from "lucide-react";
 import { buttonVariants } from "@/components/ui/button";
+import ValuationHero from "@/components/Home/ValuationHero";
+import useHomeValuation from "@/hooks/useHomeValuation";
 import { cn } from "@/lib/utils";
 
-/** Shown when both roster and socials steps are complete. */
-const RosterVerifiedPanel = () => (
-  <section className="flex flex-col items-center gap-4 text-center py-12">
-    <CheckCircle2 className="size-10 text-[#22c55e]" />
-    <div>
-      <h1 className="text-2xl font-semibold text-foreground">
-        Roster verified
-      </h1>
-      <p className="text-sm text-muted-foreground mt-1">
-        Your artists and their socials are confirmed. Reports, research, and
-        tasks will use them from here on.
-      </p>
-    </div>
-    <Link href="/" className={cn(buttonVariants(), "min-w-[200px]")}>
-      Back to Recoup
-    </Link>
-  </section>
-);
+/**
+ * Shown when both roster and socials steps are complete.
+ *
+ * Ends setup on the account's catalog valuation — the number a valuation lead
+ * converted on — rather than a generic green check (chat#1889). Reuses the
+ * homepage hero (`useHomeValuation` + `ValuationHero`) instead of building a
+ * second valuation surface. Falls back to the confirmation state when there is
+ * no valuation yet, so a cold-start account still gets a clear finish and a
+ * pointer at the step that unlocks one.
+ */
+const RosterVerifiedPanel = () => {
+  const valuation = useHomeValuation();
+
+  if (valuation.show) {
+    return (
+      <section className="flex flex-col items-center gap-4 py-8">
+        <ValuationHero
+          artistName={valuation.artistName}
+          artistImage={valuation.artistImage}
+          valuation={valuation.valuation}
+          measuredTrackCount={valuation.measuredTrackCount}
+        />
+        <p className="text-sm text-muted-foreground text-center">
+          This is your baseline. A weekly report re-measures it and emails you
+          the trend.
+        </p>
+        <Link
+          href="/setup/tasks"
+          className={cn(buttonVariants(), "min-w-[200px]")}
+        >
+          Set up your weekly report
+        </Link>
+      </section>
+    );
+  }
+
+  return (
+    <section className="flex flex-col items-center gap-4 text-center py-12">
+      <CheckCircle2 className="size-10 text-[#22c55e]" />
+      <div>
+        <h1 className="text-2xl font-semibold text-foreground">
+          Roster verified
+        </h1>
+        <p className="text-sm text-muted-foreground mt-1">
+          Your artists and their socials are confirmed. Claim your catalog to
+          see what it is worth.
+        </p>
+      </div>
+      <Link
+        href="/setup/catalog"
+        className={cn(buttonVariants(), "min-w-[200px]")}
+      >
+        Claim your catalog
+      </Link>
+    </section>
+  );
+};
 
 export default RosterVerifiedPanel;
--- a/components/Onboarding/SetupValuation.tsx
+++ b/components/Onboarding/SetupValuation.tsx
@@ -1,0 +1,65 @@
+"use client";
+
+import { useEffect } from "react";
+import Link from "next/link";
+import { useRouter } from "next/navigation";
+import { buttonVariants } from "@/components/ui/button";
+import { Skeleton } from "@/components/ui/skeleton";
+import ValuationHero from "@/components/Home/ValuationHero";
+import useCatalogs from "@/hooks/useCatalogs";
+import useHomeValuation from "@/hooks/useHomeValuation";
+import { cn } from "@/lib/utils";
+
+/**
+ * `/setup/valuation` — the welcome email's payoff link. Renders the account's
+ * baseline valuation with the shared homepage hero (chat#1889); the route used
+ * to be a bare redirect to `/catalogs`, so the email's "See your baseline
+ * valuation" link had no real destination.
+ *
+ * Falls back to `/catalogs` only when the account has no catalog to value, so a
+ * cold-start signup still lands somewhere actionable rather than on an empty
+ * hero.
+ */
+const SetupValuation = () => {
+  const router = useRouter();
+  const valuation = useHomeValuation();
+  const { data, isLoading, isError } = useCatalogs();
+  const hasCatalog = !!data?.catalogs?.length;
+
+  useEffect(() => {
+    if (isLoading) return;
+    if (!hasCatalog || isError) router.replace("/catalogs");
+  }, [isLoading, hasCatalog, isError, router]);
+
+  if (isLoading || !valuation.show) {
+    return (
+      <div className="mx-auto flex w-full max-w-xl flex-col gap-4 px-6 py-8">
+        <Skeleton className="h-8 w-1/2 rounded-lg" />
+        <Skeleton className="h-[168px] w-full rounded-xl" />
+      </div>
+    );
+  }
+
+  return (
+    <section className="mx-auto flex w-full max-w-xl flex-col items-center gap-4 px-6 py-8">
+      <ValuationHero
+        artistName={valuation.artistName}
+        artistImage={valuation.artistImage}
+        valuation={valuation.valuation}
+        measuredTrackCount={valuation.measuredTrackCount}
+      />
+      <p className="text-center text-sm text-muted-foreground">
+        This estimate goes stale the day you close the tab. A weekly report
+        re-measures your catalog and emails you the trend.
+      </p>
+      <Link
+        href="/setup/tasks"
+        className={cn(buttonVariants(), "min-w-[200px]")}
+      >
+        Set up your weekly report
+      </Link>
+    </section>
+  );
+};
+
+export default SetupValuation;
--- a/components/Onboarding/SetupValuation.tsx
+++ b/components/Onboarding/SetupValuation.tsx
@@ -23,15 +23,20 @@ import { cn } from "@/lib/utils";
 const SetupValuation = () => {
   const router = useRouter();
   const valuation = useHomeValuation();
-  const { data, isLoading, isError } = useCatalogs();
+  // `isPending`, not `isLoading`: useCatalogs is `enabled: !!accountId &&
+  // authenticated`, and a disabled TanStack Query v5 query reports
+  // isPending true / isFetching false — so isLoading is false while Privy is
+  // still resolving. Redirecting on that bounced every cold load of this
+  // route, which is exactly where the welcome email's payoff link points.
+  const { data, isPending, isError } = useCatalogs();
   const hasCatalog = !!data?.catalogs?.length;
 
   useEffect(() => {
-    if (isLoading) return;
+    if (isPending) return;
     if (!hasCatalog || isError) router.replace("/catalogs");
-  }, [isLoading, hasCatalog, isError, router]);
+  }, [isPending, hasCatalog, isError, router]);
 
-  if (isLoading || !valuation.show) {
+  if (isPending || !valuation.show) {
     return (
       <div className="mx-auto flex w-full max-w-xl flex-col gap-4 px-6 py-8">
         <Skeleton className="h-8 w-1/2 rounded-lg" />
--- a/components/Onboarding/__tests__/RosterVerifiedPanel.test.tsx
+++ b/components/Onboarding/__tests__/RosterVerifiedPanel.test.tsx
@@ -1,0 +1,52 @@
+// @vitest-environment jsdom
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import RosterVerifiedPanel from "@/components/Onboarding/RosterVerifiedPanel";
+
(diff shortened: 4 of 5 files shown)


sweetmantech:
## Preview verification — rebased onto `main`, one blocker found

Rebased off `main` (was 4 behind: #1887, #1886, #1890, #1892) and force-pushed as `402e4769`. Clean rebase; diff unchanged at **+183/−24 across 4 files**.

- **Preview:** `https://chat-jga73hv29-recoup.vercel.app` (built from `402e4769`)
- **Tests:** 82 passed / 22 files. `tsc --noEmit` → 13, identical to `main`'s baseline.
- **Account:** `sweetman+july2820260924@recoupable.com` — 2 artists, **0 catalogs**, never valued. The cold-start cohort #1892 just unblocked.

### What works

| # | Check | Result | |
|---|-------|--------|---|
| 1 | Cold-start fallback replaces the generic green check | "Your artists and their socials are confirmed. **Claim your catalog to see what it is worth.**" + `Claim your catalog` → `/setup/catalog` | ✅ |
| 2 | Valuation reward renders on roster-verified | `ValuationHero` with **$389K**, range **$267K to $547K · 79 tracks measured**, artist avatar + name | ✅ |
| 3 | Reward CTA | "Set up your weekly report" → `/setup/tasks` | ✅ |
| 4 | `/setup/valuation` with no catalog falls back | redirects to `/catalogs` | ✅ |
| 5 | Copy is em-dash free | zero `/[—–]/` in rendered text; the hero range reads "$267K **to** $547K" | ✅ |

Check 2 was driven by intercepting `/catalogs` + the measurements endpoint to return a valid catalog and band (this account has neither), so the hero path is proven to render — the numbers below are injected, not real:

valuation hero

Cold-start fallback, unmodified:

cold start

Side benefit confirmed: with a catalog present the roster copy correctly flips to "We set these artists up from your valuation", so [#1892](https://github.com/recoupable/chat/pull/1892)'s `hasValuation` keying works against a real catalog rather than only in unit tests.

---

## 🔴 Blocker — `/setup/valuation` redirects to `/catalogs` on every cold load, even with a catalog

This is the route the welcome email's "See your baseline valuation" link points at, and the reason the PR exists. It cannot render the hero when reached that way.

`SetupValuation` gates its redirect on `isLoading` from `useCatalogs`:

```ts
useEffect(() => {
  if (isLoading) return;
  if (!hasCatalog || isError) router.replace("/catalogs");
}, [isLoading, hasCatalog, isError, router]);
```

But `useCatalogs` is `enabled: !!accountId && authenticated`, and in **TanStack Query v5** (`^5.66.0`) a *disabled* query is `isPending: true` / `isFetching: false`, so **`isLoading` is `false`**. While Privy is still resolving, the guard falls through with `data === undefined`, `hasCatalog === false`, and the redirect fires.

Instrumented on the preview — patched `History.replaceState` and `fetch` to timestamp both:

```
init                              @0ms
replaceState -> /setup/valuation  @129ms
replaceState -> /catalogs         @484ms   ← redirect fires
catalogs-fetch-START              @1312ms  ← request starts 828ms LATER
catalogs-fetch-END status=200     @1851ms
```

The redirect happens **828ms before the catalogs request is even issued**. I also re-ran it with `/catalogs` intercepted to return a populated catalog: still redirected, because the decision is made before any fetch.

**Fix:** gate on `isPending` rather than `isLoading` (or on `status !== "success"`, or hold until `authenticated && accountId`). One line, but it's the difference between the feature working and never rendering from its primary entry point.

**Second, related:** once that's fixed, `hasCatalog === true` with `valuation.show === false` renders `<Skeleton>` forever — the redirect can't fire (there *is* a catalog) and the hero won't render. That's reachable for a freshly claimed catalog whose measurements haven't materialised, or when `getValuationHeroState` hides the hero because the artist-scope echo doesn't match. Worth a terminal state rather than an indefinite skeleton.

## 🟡 "Claim your catalog" moves the dead end one hop

The new cold-start CTA is the right idea, but its destination is empty. `/setup/catalog` is `redirect("/catalogs")`, and `/catalogs` for an account with no catalog renders **"No catalogs found."** with no create affordance — enumerating every `button`/`a` on the page returns only nav chrome (`Recoupable`, `New Chat`, `Agents`, `Tasks`, `Files`, account menu). There is nothing to click to claim anything.

dead end

So the flow still terminates without a payoff for the cohort it targets; it just terminates one screen later, with a promise attached. Not necessarily this PR's job to build the claim step — but as shipped, the CTA makes a promise the next screen doesn't keep. Options: point it somewhere that can actually create a catalog, or soften the copy until that step exists.

## ⚪ `/setup/catalog` repeats the `redirect()` stub pattern #1887 retired

```
GET /setup/catalog     → 200   (no Location)
GET /setup/valuation   → 200   (real page, expected)
GET /onboarding/roster → 308 → /setup/artists   (config redirect from #1887)
```

`/setup/catalog` answers **200**, not a 3xx — Next prerenders a page whose body is only `redirect()`, so the hop ships in the RSC payload and fires on hydration. That is verbatim the finding that made [#1887](https://github.com/recoupable/chat/pull/1887) move to `next.config.mjs` redirects, recorded as a decision on chat#1889 (2026-07-27). It matters here because **the welcome email links to `/setup/catalog` as step 3**, so a crawler or link previewer gets a 200 and an empty shell. Pre-existing, not introduced by this PR, but this PR makes it the destination of a primary CTA. A `next.config.mjs` rule would settle it consistently with #1887.

---

### Not verified

- The hero was rendered from **injected** catalog + measurement data; no live valuation exists on any account I can sign into. The real numbers path is unchanged from the homepage hero this reuses, but it has not been exercised end-to-end here.
- `isError` on the catalogs query was not exercised separately from `!hasCatalog`.

### Recommendation

The reward itself is good and the fallback is the right shape. I'd fix the 🔴 before merge — it's one line and without it the email's payoff link never reaches the feature — and make a call on the 🟡, since a CTA that lands on "No catalogs found" reads worse than the green check it replaced. Verified on `402e4769`.

sweetmantech:
## 🔴 Blocker fixed — `61911a6a`

`SetupValuation` now gates on **`isPending`** rather than `isLoading`, in both the redirect effect and the skeleton branch.

```diff
-  const { data, isLoading, isError } = useCatalogs();
+  const { data, isPending, isError } = useCatalogs();
   const hasCatalog = !!data?.catalogs?.length;

   useEffect(() => {
-    if (isLoading) return;
+    if (isPending) return;
     if (!hasCatalog || isError) router.replace("/catalogs");
-  }, [isLoading, hasCatalog, isError, router]);
+  }, [isPending, hasCatalog, isError, router]);
```

New `components/Onboarding/__tests__/SetupValuation.test.tsx` — the component had no test file — with four cases. RED before the fix, on the one that matters:

```
FAIL  SetupValuation > does not redirect while the catalogs query has not resolved
AssertionError: expected "replace" to not be called at all, but actually been called 1 times
  1st vi.fn() call: [ "/catalogs" ]
```

| Test | Asserts |
|---|---|
| **does not redirect while the catalogs query has not resolved** | `isPending: true`, `isLoading: false` ⇒ `router.replace` **not** called |
| redirects once settled with no catalog | `isPending: false`, `catalogs: []` ⇒ `replace("/catalogs")` |
| redirects when the catalogs read fails | `isError: true` ⇒ `replace("/catalogs")` |
| renders the hero when a valuation is available | hero rendered, **no** redirect |

The last three guard the fix from over-correcting into "never redirects".

### Verified on preview `chat-akkmnwyyn-recoup.vercel.app` (built from `61911a6a`)

Same instrumentation as the original report — `History.replaceState` and `fetch` timestamped on a cold load.

**With a catalog** (intercepted, since no account I can reach has one) — previously redirected 100% of the time:

```
init                       @0ms
replaceState -> /setup/valuation  @1336ms
catalogs-fetch             @5768ms
measurements-fetch         @5777ms
measurements-fetch         @7086ms
```

**No `replaceState -> /catalogs` at all.** Final URL stays `/setup/valuation`, and the hero renders with its route-specific copy:

setup/valuation renders

**Without a catalog** (this account's real state) — the fallback still works, and now happens in the right order:

```
init                        @0ms
replaceState -> /setup/valuation   @4257ms
catalogs-fetch-START        @10138ms
catalogs-fetch-END 200      @10776ms
replaceState -> /catalogs   @16476ms   ← after the read resolves
```

Compare with `402e4769`, where the redirect fired at **484ms** and the fetch didn't start until **1312ms**. The decision is now made on data instead of on an unresolved query.

Full suite: **86 passed / 23 files**. `tsc --noEmit` → 13, unchanged from `main`. `eslint` clean.

### Still open on this PR

The 🟡 **"Claim your catalog" dead end** is unchanged — see the recommendation in the next comment. The ⚪ `/setup/catalog` returning 200 instead of a 3xx is also unchanged and is pre-existing.

The secondary skeleton risk I flagged (`hasCatalog === true` + `valuation.show === false` ⇒ indefinite skeleton) is **still present** — `isPending` fixes the auth race but not this case, which is reachable while a freshly created catalog is still being measured. Not a blocker, but worth a terminal state.

sweetmantech:
## Recommendation on 🟡 "Claim your catalog" — don't fix the copy, seed the catalog

Digging into why this account has two Spotify-id'd artists and still no catalog turned up the actual cause, and it isn't in this PR.

**A hook that already does the right thing exists.** [`hooks/useAddSpotifyArtist.ts`](https://github.com/recoupable/chat/blob/main/hooks/useAddSpotifyArtist.ts) adds a Spotify-searched artist and then, **only when the account has no catalog yet**, fire-and-forget kicks `POST /api/valuation` for that artist's Spotify id via [`lib/valuation/runValuation.ts`](https://github.com/recoupable/chat/blob/main/lib/valuation/runValuation.ts) — which materialises an account-owned catalog and value band. Its docstring states the intent outright:

> seeding the onboarding catalog in the background so the "Claim your catalog" step is already complete by the time the user reaches it (chat#1867)

**The onboarding form doesn't use it.** [chat#1892](https://github.com/recoupable/chat/pull/1892) wired `AddArtistForm` to `useAddRosterArtist`, which creates the artist and enriches it but never runs a valuation. And the enrichment it added is a re-implementation: [`lib/artists/addSpotifyArtist.ts`](https://github.com/recoupable/chat/blob/main/lib/artists/addSpotifyArtist.ts) already did the identical `createRosterArtist` → `saveArtist({ image, profileUrls: { SPOTIFY } })` sequence, and predates it.

| | `useAddSpotifyArtist` (`/artists` dialog) | `useAddRosterArtist` (onboarding, post-#1892) |
|---|---|---|
| Create by name | ✅ | ✅ |
| Attach Spotify URL + avatar | ✅ (`addSpotifyArtist`) | ✅ (re-implemented inline) |
| Select + toast the new artist | ✅ | — |
| **Seed the catalog via `runValuation`** | ✅ when `catalogs.length === 0` | ❌ |

So an artist added through `/artists` gets a catalog; the same artist added through `/setup/artists` does not. That asymmetry is why the cold-start branch of this PR is reachable at all, and why its CTA has nowhere to go. `useAddRosterArtist` has exactly one caller (`AddArtistForm`), so this is a swap, not a migration.

### What I'd do

**Point `AddArtistForm` at `useAddSpotifyArtist` and delete the duplicated path.** Then a cold-start account's catalog is already materialising while they walk verify-socials (~20s, fire-and-forget), and by the time they hit the end of setup **this PR's `ValuationHero` is the real ending** rather than the fallback. That is strictly better than making the fallback's CTA work: it removes the dead end instead of furnishing it, deletes code rather than adding it, and it is what #1867 already intended.

**⚠️ One thing must come with it.** `useAddSpotifyArtist` has the *same* fatal-enrichment bug [#1892](https://github.com/recoupable/chat/pull/1892) just fixed — `addSpotifyArtist` awaits `saveArtist` and a throw propagates to the hook's shared `catch`, so a failed `PATCH` reports "Failed to add artist" over an artist that was already created, and retrying duplicates it. Swapping without carrying that `try`/`catch` into `addSpotifyArtist` would reintroduce the defect. It should also be pinned by a test, as it now is on `useAddRosterArtist`.

### Sequencing

This is its own change, not something to fold into this PR — it touches the add path, not the reward surface, and it deserves its own preview verification. This PR's fallback stays correct and useful either way: it is the right thing to show when there genuinely is no catalog (valuation failed, or an account that predates seeding).

**Interim:** if the seeding change doesn't land soon, soften this PR's CTA copy, because "Claim your catalog to see what it is worth" → "No catalogs found" with no create affordance reads worse than the generic green check it replaced. But that is a stopgap; the swap is the fix.

sweetmantech:
## Final verification — all four surfaces re-tested on `61911a6a`

The earlier screenshots were taken on `402e4769`, **before** the `isPending` fix. Re-running every surface this PR touches against the final commit so nothing on this PR is evidence from a superseded build.

- **Commit:** `61911a6a` (PR head)
- **Preview:** `https://chat-akkmnwyyn-recoup.vercel.app`
- **Checks:** unit tests ✅ · CodeRabbit ✅ · cubic ✅ · Vercel ✅
- **Account:** `sweetman+july2820260924@recoupable.com` — 2 artists, 0 catalogs, never valued

Two of the four states need a catalog this account doesn't have, so those were driven by intercepting `/catalogs` + the measurements endpoint. **The numbers in C and D are injected, not real.** A and B are the account's genuine state, unmodified.

| # | Surface | State | Result | |
|---|---|---|---|---|
| A | `/setup/valuation` | no catalog (real) | falls back to `/catalogs` **after** the read resolves | ✅ |
| B | `RosterVerifiedPanel` | no valuation (real) | "Claim your catalog to see what it is worth." + CTA | ✅ |
| C | `RosterVerifiedPanel` | valuation present | `ValuationHero` + "Set up your weekly report" | ✅ |
| D | `/setup/valuation` | catalog present, **cold load** | renders the hero, **no redirect** | ✅ |

### D — the fix, on the final commit

This is the case that was broken. Cold load, catalog present:

```
init                              @0ms
replaceState -> /setup/valuation  @122ms
catalogs-fetch                    @1492ms
```

No `replaceState -> /catalogs` anywhere in the trace, and `location.pathname === '/setup/valuation'` at the end. On `402e4769` this same load redirected at **484ms**, before the catalogs request was even issued.

setup/valuation renders

### A — the fallback still fires, and now on data

Same route, this account's real state (no catalog). Lands on `/catalogs`, so the fix didn't over-correct into "never redirects" — the three guard tests in the suite cover the same property.

no catalog fallback

### B — cold-start completion panel

Walked the real flow: `/setup/artists` → continue → `/setup/socials` → continue. Replaces the old "Back to Recoup" green check.

cold start panel

### C — the reward, when a valuation exists

Same walk with a catalog + band injected. `ValuationHero` ends the flow instead of the check.

valuation hero panel

Also asserted on D: **zero em dashes** in the rendered text (the hero range reads "$267K **to** $547K").

### Unchanged from the earlier report

- 🟡 **"Claim your catalog" still dead-ends** at "No catalogs found" with no create affordance. Recommendation is in [this comment](https://github.com/recoupable/chat/pull/1895#issuecomment-5108235684) and is now tracked as **row 8** on chat#1889 — the fix is to seed the catalog on first artist add via `useAddSpotifyArtist`, not to reword this CTA.
- ⚪ `/setup/catalog` still answers **200** rather than a 3xx (`redirect()` page stub, the pattern #1887 retired). Pre-existing.
- The `hasCatalog === true` + `valuation.show === false` indefinite-skeleton case is still reachable; `isPending` fixed the auth race, not this one.

Suite at `61911a6a`: **86 passed / 23 files**, `tsc --noEmit` → 13 (identical to `main`), eslint clean.

PR #1900, opened 2026-07-28
feat(onboarding): seed the catalog when the first artist is added in setup (chat#1889 row 8)

sweetmantech:
Matrix row 8 in chat#1889 — the fix for the dead end [#1895](https://github.com/recoupable/chat/pull/1895) surfaced.

## Why

`/setup/artists` added artists through `useAddRosterArtist`, which creates and enriches but **never values**. [`useAddSpotifyArtist`](https://github.com/recoupable/chat/blob/main/hooks/useAddSpotifyArtist.ts) — already used by the `/artists` dialog — also fire-and-forget kicks `POST /api/valuation` via `runValuation` when the account has no catalog yet. Its docstring has said so since chat#1867:

> seeding the onboarding catalog in the background so the "Claim your catalog" step is already complete by the time the user reaches it

So the same artist added at `/artists` got a catalog and at `/setup/artists` did not. Verified live on a fresh account during #1895's review: two artists added through setup, both carrying real Spotify ids, `GET /api/accounts/{id}/catalogs` still `[]`. With #1895 on prod that means every cold-start account now finishes setup on the **"Claim your catalog"** fallback, whose CTA lands on `/catalogs` → "No catalogs found." with no create affordance.

## What changed

- **`AddArtistForm`** adds through `useAddSpotifyArtist`, passing the whole Spotify result.
- **`useAddRosterArtist` deleted** (and its test). Its enrichment duplicated [`addSpotifyArtist`](https://github.com/recoupable/chat/blob/main/lib/artists/addSpotifyArtist.ts) — identical `createRosterArtist` → `saveArtist({ image, profileUrls: { SPOTIFY } })` — which predates chat#1892. One caller, so this is a swap that removes code.
- **`addSpotifyArtist` enrichment is now non-fatal**, carrying chat#1892's guard across. It awaited `saveArtist` bare, so a failed `PATCH` threw past the create that had already succeeded: the caller reported failure over an existing artist and a retry duplicated it. **This had to land with the swap or the swap would have reintroduced the defect chat#1892 just closed.**

No api or docs change: this reuses `POST /api/valuation` (api#776) exactly as the `/artists` dialog already does, so there is no contract to write first.

## Verification

TDD, red → green, two units:

```
# RED 1 — enrichment failure propagates
FAIL lib/artists/__tests__/addSpotifyArtist.test.ts > returns the created artist when enrichment fails
Error: Forbidden

# RED 2 — form still bound to the non-seeding hook
FAIL components/Onboarding/__tests__/AddArtistForm.test.tsx (2 failed)

# GREEN
pnpm exec vitest run components lib hooks
  Test Files  72 passed (72)
       Tests  271 passed (271)
```

`pnpm exec tsc --noEmit` → 13 errors, identical to `main`'s baseline. `eslint` clean on all touched files.

Preview verification to follow on this PR.

Tracked in chat#1889 (matrix row 8).

---
## Summary by cubic
Seeds the catalog when the first artist is added during setup by routing `AddArtistForm` through `useAddSpotifyArtist`, and adds a measuring state in `/setup/valuation` so onboarding no longer dead-ends (chat#1889 row 8). This mirrors the `/artists` dialog and kicks `POST /api/valuation` on first add.

- **Bug Fixes**
  - Seeding: `useAddSpotifyArtist` now runs valuation on the first artist added in setup, preventing empty-catalog onboarding paths.
  - Measuring: `/setup/valuation` shows a “Measuring your catalog” state when a catalog exists but has no valuation yet, avoiding an endless skeleton and guiding users to `/setup/tasks`.

- **Refactors**
  - Removed `useAddRosterArtist`; `AddArtistForm` now calls `useAddSpotifyArtist` with the full Spotify result. Tests updated.
  - Extracted `MeasuringCatalogPanel` into its own component and used it in `SetupValuation`.

<sup>Written for commit e86f3e74f9fe8132b2968af45a7c4f8993b9e174. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1900?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Onboarding now adds artists directly from Spotify selections, bringing in profile details and avatar when available.
  - Added an intermediate **“Measuring your catalog”** screen for `/setup/valuation` when measurements are still pending, with a link to setup tasks.
- **Bug Fixes**
  - Artist addition no longer fails if Spotify enrichment (profile/avatar) temporarily errors; the base artist is still created for later repair.
- **UI Improvements**
  - Updated valuation setup rendering so loading vs “not yet showing” states display the correct screens (skeleton only during loading, measuring panel otherwise).


sweetmantech:
## Preview verification — the seeding works end to end, and it surfaced one thing worth fixing here

Verified on `chat-qg3k0lgqt-recoup.vercel.app` (built from `c7d544d6`) using the same cold-start account as the #1892/#1895 passes: `sweetman+july2820260924@recoupable.com`, which had **2 artists and `catalogs: []`** — the exact state that proved the gap.

### The call sequence now includes the seeding

Added *Del Water Gap* through `/setup/artists` with `fetch` instrumented:

```
GET   /api/artists?
POST  /api/artists                                  ← create
PATCH /api/artists/289d8342-543d-4212-9c44-70f3e6ffdf83   ← enrich (Spotify URL + avatar)
GET   /api/artists?                                 ← roster refresh
POST  /api/valuation                                ← the seeding. Never fired from setup before.
```

Toast confirmed: **"Valuing Del Water Gap's catalog…"**

valuation kicked

### The catalog materialises, and the dead end is gone

| Check | Before this PR | After |
|---|---|---|
| `GET /api/accounts/{id}/catalogs` | `[]` after two artists added in setup | **1 catalog** — "Valuation Catalog" |
| `/catalogs` (where "Claim your catalog" lands) | "No catalogs found." + no create affordance | **"Valuation Catalog · 67 songs · Created: 7/28/2026"** |

catalog exists

That is the Done-when for row 8: the CTA that #1895 added now lands somewhere real.

### 🔴 What this surfaced, and fixed in `0c889248`

Immediately after the seeded valuation returned, `/setup/valuation` rendered **a skeleton with no exit**: the redirect can't fire (a catalog now exists) and the hero can't render (measurements haven't landed). Reproduced live, not theorised:

blank measuring window

I flagged this as a possible risk when reviewing #1895 and judged it non-blocking because it needed a catalog with no measurements — a state nothing produced routinely. **Row 8 produces it on every cold-start signup**, for the ~minute between catalog creation and measurement capture. So it is fixed here rather than left: `SetupValuation` now renders a terminal *"Measuring your catalog"* state with a "Set up your weekly report" exit, instead of an indefinite skeleton.

### Suite

```
pnpm exec vitest run components lib hooks
  Test Files  72 passed (72)
       Tests  272 passed (272)
```

`tsc --noEmit` → 13, identical to `main`. `eslint` clean.

### Not verified

- **The new "Measuring your catalog" state is unit-tested (RED→GREEN) but not browser-confirmed.** The preview Privy session expired before I could re-test `0c889248`, and re-authenticating needs an email OTP I can't complete unattended. The bug it fixes *is* browser-confirmed (screenshot above); the fix itself is not. Worth a click-through before merge.
- The end-of-setup panel still showed the "Claim your catalog" fallback immediately after seeding, for the same measurement-lag reason. Expected, and it self-resolves once measurements land — but it means a user who races through setup in under a minute still won't see the hero on their first pass.

components/Onboarding/SetupValuation.tsx
```diff
@@ -45,6 +45,31 @@ const SetupValuation = () => {
...
+          We are pulling play counts for every track. This usually takes a
+          minute. Your baseline value appears here as soon as it is ready.
+        </p>
+        <Link
+          href="/setup/tasks"
+          className={cn(buttonVariants(), "min-w-[200px]")}
+        >
+          Set up your weekly report
+        </Link>
+      </section>
```
sweetmantech: OCP
- actual: net new code defined inline of existing component file
- required: new component file for the new component code definition.

sweetmantech: Done in `2f6aee00` — extracted to `components/Onboarding/MeasuringCatalogPanel.tsx` (one exported component per file). `SetupValuation` now composes it (`if (!valuation.show) return <MeasuringCatalogPanel />;`) instead of carrying the markup inline, so a future state is a new file rather than another branch edited into this one. 89 tests pass; tsc at baseline; eslint clean.

sweetmantech:
## 🔴 Blocker found on re-test — seeding creates a **duplicate roster artist**. Moving this to draft.

Caught by @sweetmantech spotting the same artist twice in the sidebar of my own screenshot. It is real, and it is caused by this PR.

Re-tested from a genuinely empty account (`sweetman+july2820261713@…`, brand new, zero artists, zero catalogs). Added **one** artist. `GET /api/artists` afterwards:

```json
[
  { "name": "Del Water Gap", "account_id": "6e4d3f5f-87d5-47f3-a9b6-b7058d24bb99",
    "socials": ["open.spotify.com/artist/0xPoVNPnxIIUS1vrxAYV00"] },
  { "name": "Del Water Gap", "account_id": "7647f901-a640-434c-b9c9-a18682435092",
    "socials": ["open.spotify.com/artist/0xPoVNPnxIIUS1vrxAYV00"] }
]
```

Same name, same avatar, same Spotify id, two roster entries.

### Mechanism

The instrumented call log shows the client issued **exactly one** `POST /api/artists`:

```
GET   /api/artists?
POST  /api/artists                            ← one create
PATCH /api/artists/6e4d3f5f-…                 ← enrich
GET   /api/artists?
POST  /api/valuation                          ← seeding
```

So the second artist was created **server-side, by the valuation**. [`runValuationHandler`](https://github.com/recoupable/api/blob/main/lib/valuation/runValuationHandler.ts) step 4 runs `createSnapshotCatalog` → [`attachCanonicalArtistToAccount`](https://github.com/recoupable/api/blob/main/lib/catalog/attachCanonicalArtistToAccount.ts), which resolves the artist from the songs graph (ISRC → `song_artists` → dominant artist) and links **that** account — a different row from the one this form just minted. Its own docstring names the failure mode:

> This replaces the marketing funnel's **per-signup artist creation — which minted duplicate, song-less roster artists** — as the roster source of truth for funnel signups.

The api deliberately moved away from creating an artist per signup *because it duplicated*. This PR pairs a client-side create with a server-side canonical attach and reintroduces exactly that.

### Scope

**Pre-existing on the `/artists` dialog path**, not new code: `useAddSpotifyArtist` has always done create-then-seed, so any account whose *first* artist was added through `AddArtistDialog` gets the same duplicate. What this PR changes is that it becomes **the default path for every cold-start signup** — so it goes from an edge case to universal. It must not merge as-is.

### Options

1. **Reuse instead of insert (api-side, preferred).** `POST /api/valuation` links the account's existing artist when one already carries the same Spotify id, rather than attaching a second. Fixes the dialog path too, and matches `attachCanonicalArtistToAccount`'s stated intent.
2. **Don't pre-create in onboarding.** Let the valuation's canonical attach populate the roster. Cleaner conceptually, but the roster step needs an artist to enable Continue, so the user would stare at an empty list for ~20s.
3. **Reconcile client-side after seeding.** Drop the locally-created artist once the canonical one appears. Hides the problem rather than fixing it, and races.

I'd take (1). It is the only one that also fixes the existing dialog path, and it belongs in `api` where the duplicate is actually created.

### What still stands from this PR

Everything else verified on the new account: seeding fires on the first add, the catalog materialises, and the **"Measuring your catalog"** terminal state renders correctly in the real window with a working exit — the case that was a blank skeleton before. Screenshots to follow. The duplicate is the one blocker.

sweetmantech:
Blocker fix is up: **[api#791](https://github.com/recoupable/api/pull/791)** — `linkSearchedArtistToAccount` now reuses the canonical artist for a Spotify id instead of calling `createArtistInDb` unconditionally.

Per @sweetmantech's call the lookup is **global, not per-account**: artists are canonical and shared and `account_artist_ids` is the join that lets many accounts roster one artist (chat#1866), so a per-account dedup would still have given every account its own copy — it would only have stopped one account duplicating itself.

Merge order is **api#791 → this PR**. Once #791 is on `main` I'll re-run the cold-start walk on a fresh account and confirm one artist, one catalog, then take this out of draft.

sweetmantech:
## Final verification on `e86f3e74` — the arc closes, with real numbers

Rebased onto latest `main` (post-#1892/#1895/#1901; the superseded `addSpotifyArtist` guard resolved **to main** — an inverted first resolution was caught by #1901's own tests and corrected in the same push). Remaining scope: the onboarding form swap to `useAddSpotifyArtist`, `useAddRosterArtist` deleted, and the `MeasuringCatalogPanel` terminal state. 273 tests pass; `tsc` at **8** (below main's 13 — deletions removed baseline errors).

**Preview walk** on `chat-f260o93oj` (sha-verified), brand-new account `sweetman+july2920261250@…`, against `test-recoup-api` synced through api#794 minutes earlier.

### The instrumented add — onboarding now speaks the full canonical contract

```
POST /api/artists   {"name":"Del Water Gap","spotify_artist_id":"0xPoVNPnxIIUS1vrxAYV00"}  → 200
POST /api/valuation {"spotify_artist_id":"0xPoVNPnxIIUS1vrxAYV00"}                          (seeding)
```

**200, not 201** — the cold-start signup rostered the *cross-account canonical* (`7647f901`, now shared by 13 accounts) instead of minting a row. **Zero `profileUrls` PATCHes** in the whole log. This closes the last regrowth vector.

roster

### Post-completion assertion — the one that raced on 07-28, now taken after the valuation finished

| Check | Result |
|---|---|
| Roster count **after** valuation completion | **1** |
| That artist is the canonical | ✅ `7647f901` |
| DWG artist rows globally (SQL) | **1** — the walk minted nothing |
| Spotify ids resolving to >1 artist, globally | **0** |

### The payoff, with genuine data

`/setup/valuation` rendered the real `ValuationHero` — **$2.1M, range $1.4M–$2.9M, 67 tracks measured** — Del Water Gap's actual numbers, no interception. The canonical's songs were already measured, so the flow skipped the measuring window entirely; the `MeasuringCatalogPanel` (browser-verified live on the earlier `…2156` walk, [screenshot](https://raw.githubusercontent.com/recoupable/chat/assets/pr1900/shots/02-measuring-panel.png)) is now the fallback for genuinely-new catalogs only.

hero

A cold-start signup now goes: search → one shared canonical → seeded valuation → **their catalog's real value ends the flow**. Ready to merge — closes row 11 and the rows 8–11 arc on chat#1889.

Issue #1912, opened 2026-07-30
Ben's first-session blockers: the catalog page dead-ends, the side menu contradicts itself, and nothing links the tooling (all reproduced 2026-07-30)


sweetmantech:
Tracking issue for the three defects a live outreach customer hit in one sitting on 2026-07-29 — account `148e1644`, first outreach cohort. Migrated out of [chat#1902](https://github.com/recoupable/chat/issues/1902) (the web UX conversion audit) 2026-07-30 so a single named customer's blocked path has one owner instead of three rows scattered across two priority buckets.

**Status 2026-07-30: rows 1-5 are merged and preview-verified — but do not yet tell the reporter this is fixed.** His first sentence ("the information wasn't loading") most likely describes the marketing valuation form, which is **row 8** and still open. His second (the referral's side-menu trouble) was never confirmed against what that person actually saw. And his third ("waiting on the newsletter") resolves on its own: his weekly task was created Monday 2026-07-27 at 19:56, after that morning's 09:00 slot, so his first report is due **Monday 2026-08-03**. The three defects as filed are shipped, along with two more found while reproducing them (rows 4 and 5). Rows 1-5 are merged and preview-verified. **Rows 1-8 are shipped and preview-verified.** Rows 9 and 10 were opened 2026-07-31 from what the verification work turned up. **Row 9 shipped**; row 10 is copy-only and is the last one open. Neither blocks reporting back to the reporter. Rows 1-5 came from the reporter's three sentences plus two defects found while reproducing them; rows 6-8 were split out of merged rows rather than dropped. Row 8 landed last and is the one that most likely explains his first sentence, since he ran his valuation from marketing.

> **Why this is its own issue.** These are not three unrelated UX papercuts. They are one customer's continuous path: ran a valuation and saw nothing, re-ran the entire flow, referred someone who hit a second bug, then gave up on the UI and waited for an email. Fixing them as a set closes the loop with a real, funded, hand-outreached account that is stalled right now. #1902 remains the tracker for the pricing→trial funnel; this one is the post-signup path.

> **Every row below was reproduced on prod 2026-07-30** on a purpose-built account configured to match Ben's exactly (see **Recreation**). Three of the claims we had written down were **wrong**, and one row's stated root cause was **inverted** — details in each row and in **Corrections**. Screenshots are customer-perspective, hosted on the [`assets/issue1912`](https://github.com/recoupable/chat/tree/assets/issue1912/shots) branch.

[...]

PR #1914, opened 2026-07-30
fix(nav): the side menu stops contradicting its own auth state (chat#1912 row 2)

sweetmantech:
Closes row 2 of [chat#1912](https://github.com/recoupable/chat/issues/1912).

## Why

Reproduced on prod 2026-07-30 at 390px, **signed out**: the opened side menu shows a **Sign In** button at the top and, at the bottom of the same panel, the user-chip slot rendering a **permanently loading skeleton** (accessible name `Loading user profile`, `disabled`). Still both present **12 seconds later** — a first-time visitor is told to sign in and simultaneously shown a profile chip loading forever for an account that does not exist.

Root cause, [`UserProfileButton.tsx`](https://github.com/recoupable/chat/blob/main/components/Sidebar/UserProfileButton.tsx):

```ts
if (!userData) return <UserProfileButtonSkeleton />;
```

`userData` is absent in two very different situations — "authenticated, account still fetching" and "nobody is signed in" — and both rendered the skeleton. For the signed-out case nothing ever arrives to resolve it.

Originally reported by a referred first-time visitor, relayed by a live outreach customer on 2026-07-29 ("I put someone else into your site and they had issues with the side menu").

## What changed

- New `lib/auth/getUserProfileState.ts` — a pure `loading | signed-out | ready` decision, gated on Privy's `authenticated` flag rather than on the presence of account data.
- `UserProfileButton` renders **nothing** when signed out, the skeleton only while genuinely loading, the chip when ready.

Gating on `authenticated` rather than `!!userData` also fixes the logout path: the chip disappears the moment the session goes, instead of lingering until the account query is evicted. There is a test for that.

## Scope note

The issue also recorded a **signed-in** variant of this bug (Sign In alongside the authenticated avatar chip, plus a Privy `already logged in` warning). That state **did not reproduce** across normal load, Slow 3G + 4x CPU throttling, and force-opening the menu on first paint — details in the issue. This PR fixes the variant that does reproduce. The original sighting is un-reproduced, not disproven, so row 2 should be re-checked against the preview before it is closed.

## Tests

TDD red→green, two files:

- `lib/auth/__tests__/getUserProfileState.test.ts` — 5 cases (RED: module not found), including the stale-data-after-logout guard.
- `components/Sidebar/__tests__/UserProfileButton.test.tsx` — 4 cases. The signed-out case was **RED against the real defect** (`expected <button …> to be null`), the other three green throughout, proving the loading and ready paths are unchanged.

Full suite **291 passed / 77 files**, `tsc --noEmit` clean.

## Verification

Preview verification against the issue's **Works when** script (390px signed-out, the 15s re-check, signed-in, console) will be posted as a PR comment once the preview builds.

---
## Summary by cubic
Fixes the side menu so signed-out users no longer see a stuck profile skeleton next to Sign In, the profile chip disappears immediately on logout, and the account slot now offers Sign in. Wallet-only users (e.g. mini-app) keep their profile chip. Addresses chat#1912 row 2.

- **Bug Fixes**
  - Updated `lib/auth/getUserProfileState` to decide `loading | signed-out | ready` using `@privy-io/react-auth` readiness/auth plus wallet presence, not `userData`. Treats wallet-only sessions as signed in and marks logout as signed-out immediately.
  - Updated `UserProfileButton` to render `SignInSlotButton` when signed out, show a skeleton only while loading, and keep the chip for wallet-only users.
  - Added tests for the helper and component covering signed-out, loading, ready, logout, wallet-only, and click-to-login.

<sup>Written for commit f140e5d15f0986a54b66295c4b78f3e7f75bc559. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/components/Sidebar/UserProfileButton.tsx
+++ b/components/Sidebar/UserProfileButton.tsx
@@ -1,6 +1,8 @@
+import { usePrivy } from "@privy-io/react-auth";
 import { useUserProvider } from "@/providers/UserProvder";
 import { useOrganization } from "@/providers/OrganizationProvider";
 import useAccountOrganizations from "@/hooks/useAccountOrganizations";
+import { getUserProfileState } from "@/lib/auth/getUserProfileState";
 import UserProfileDropdown from "./UserProfileDropdown";
 import UserProfileButtonSkeleton from "./UserProfileButtonSkeleton";
 import {
@@ -13,10 +15,21 @@ import { cn } from "@/lib/utils";
 
 const UserProfileButton = ({ isExpanded = true }: { isExpanded?: boolean }) => {
   const { email, userData } = useUserProvider();
+  const { ready, authenticated } = usePrivy();
   const { selectedOrgId } = useOrganization();
   const { data: organizations } = useAccountOrganizations();
 
-  if (!userData) return <UserProfileButtonSkeleton />;
+  const profileState = getUserProfileState({
+    isPrivyReady: ready,
+    isAuthenticated: authenticated,
+    hasUserData: !!userData,
+  });
+
+  // A signed-out visitor has no profile to load, so the slot stays empty
+  // rather than showing a skeleton that never resolves next to a Sign In
+  // button (chat#1912 row 2).
+  if (profileState === "signed-out") return null;
+  if (profileState === "loading") return <UserProfileButtonSkeleton />;
 
   const userName = userData?.name || email || userData?.wallet || "";
   const userImage = userData?.image;
--- a/components/Sidebar/UserProfileButton.tsx
+++ b/components/Sidebar/UserProfileButton.tsx
@@ -5,6 +5,7 @@ import useAccountOrganizations from "@/hooks/useAccountOrganizations";
 import { getUserProfileState } from "@/lib/auth/getUserProfileState";
 import UserProfileDropdown from "./UserProfileDropdown";
 import UserProfileButtonSkeleton from "./UserProfileButtonSkeleton";
+import SignInSlotButton from "./SignInSlotButton";
 import {
   DropdownMenu,
   DropdownMenuTrigger,
@@ -14,7 +15,7 @@ import { ChevronDown } from "lucide-react";
 import { cn } from "@/lib/utils";
 
 const UserProfileButton = ({ isExpanded = true }: { isExpanded?: boolean }) => {
-  const { email, userData } = useUserProvider();
+  const { email, userData, address, login } = useUserProvider();
   const { ready, authenticated } = usePrivy();
   const { selectedOrgId } = useOrganization();
   const { data: organizations } = useAccountOrganizations();
@@ -22,13 +23,13 @@ const UserProfileButton = ({ isExpanded = true }: { isExpanded?: boolean }) => {
   const profileState = getUserProfileState({
     isPrivyReady: ready,
     isAuthenticated: authenticated,
+    hasWallet: !!address,
     hasUserData: !!userData,
   });
 
-  // A signed-out visitor has no profile to load, so the slot stays empty
-  // rather than showing a skeleton that never resolves next to a Sign In
-  // button (chat#1912 row 2).
-  if (profileState === "signed-out") return null;
+  // A signed-out visitor has no profile to load, so the slot offers the way in
+  // rather than a skeleton that can never resolve (chat#1912 row 2).
+  if (profileState === "signed-out") return <SignInSlotButton onClick={login} />;
   if (profileState === "loading") return <UserProfileButtonSkeleton />;
 
   const userName = userData?.name || email || userData?.wallet || "";
@@ -46,8 +47,6 @@ const UserProfileButton = ({ isExpanded = true }: { isExpanded?: boolean }) => {
   const primaryName = isOrgSelected
     ? selectedOrg.organization_name || "Organization"
     : userName;
-
-  const secondaryName = isOrgSelected ? userName : "Personal";
 
   const avatarImage = isOrgSelected
     ? selectedOrg.organization_image
--- a/components/Sidebar/SignInSlotButton.tsx
+++ b/components/Sidebar/SignInSlotButton.tsx
@@ -1,0 +1,24 @@
+import { LogIn } from "lucide-react";
+import { Button } from "@/components/ui/button";
+
+/**
+ * The sidebar's account slot for a visitor who is not signed in. The slot is
+ * where account state lives, so it offers the way in rather than rendering a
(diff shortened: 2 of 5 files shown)


components/Sidebar/UserProfileButton.tsx
```diff
@@ -13,10 +15,21 @@ import { cn } from "@/lib/utils";
...
+  const profileState = getUserProfileState({
+    isPrivyReady: ready,
+    isAuthenticated: authenticated,
+    hasUserData: !!userData,
+  });
+
+  // A signed-out visitor has no profile to load, so the slot stays empty
+  // rather than showing a skeleton that never resolves next to a Sign In
+  // button (chat#1912 row 2).
+  if (profileState === "signed-out") return null;
```
sweetmantech: Instead of returning null, what do you think of returning a signin button to reinforce that as an important first step for using the chat app?

sweetmantech: Done in `f140e5d1` — the slot now renders a **Sign in** button (`SignInSlotButton`, avatar-shaped so it sits where the account chip will), wired to Privy `login`. Two tests cover it: the signed-out slot renders sign-in and never a skeleton, and clicking it opens the login prompt.

One thing worth your call, since it changes what the panel looks like signed out. **Both menus already show a Sign In affordance at the top** when signed out:

- desktop `Menu` → `NewChatButton` renders the label `Sign In` when there is no `email`
- mobile `SideMenu` → the top button renders `Sign In` when there is no `address`

So a signed-out panel now has **two** sign-in affordances, top and bottom.

I do not think that reintroduces the bug this PR fixes: the original defect was a *contradiction* (a Sign In button above a signed-in user's chip). Two buttons that both say "sign in" to a signed-out visitor are consistent, just redundant, and redundancy is arguably the point when it is the first step you want reinforced.

If you would rather have exactly one, the cleaner split is to let the account slot own sign-in and stop the top button doubling as it (it would read `New Chat` always, and still opens login because `isPrepared()` calls `login()` when there is no address, so behaviour would not change). I did not make that call unilaterally since it edits copy outside this PR's scope. Say the word and it is a two-line change.

sweetmantech:

Preview **https://chat-opvxaf8pg-recoup.vercel.app**, confirmed built from head `f140e5d1` (deployment resolved by sha, not by branch). Signed in as `sweetman@recoupable.com`.

### Results

| # | Check | Documented / before | Actual on preview | |
|---|---|---|---|---|
| 1 | Signed out, 390px, menu open | **"Sign In"** at the top **and** a grey profile skeleton pinned at the bottom (`Loading user profile`, `disabled`) | top affordance plus a **"Sign in" button** in the account slot; **no skeleton** | ✅ |
| 2 | It stays fixed — re-check after **15s** | both still present, permanently | button list **byte-identical** after 15s; `stillLoadingSkeleton: false` | ✅ |
| 3 | Signed in, 390px | n/a | account chip `sweetman@recoupable.c…` renders; `hasAnySignIn: false`, `hasSkeleton: false` | ✅ |
| 4 | Console in the signed-in session | `Attempted to log in, but user is already logged in` | **no errors or warnings at all** | ✅ |
| 5 | Signed in, 1440px | n/a | `hasUserMenu: true`, `hasSignIn: false`, `hasSkeleton: false` | ✅ |
| 6 | Signed out, 1440px | skeleton | `hasSkeleton: false`, sign-in slot present | ✅ |
| 7 | Account slot CTA | n/a | opens the Privy prompt in place, completing login through it lands signed in | ✅ |

Assertions were made programmatically against the accessible names, not read off the screenshots.

### Screenshots

**Signed out, 390px** — the state a referred first-time visitor hit. The skeleton is gone and the slot offers the way in:

signed out 390

**Signed in, 390px** — chip renders, no Sign In affordance anywhere in the panel:

signed in 390

**Signed out, 1440px**:

signed out 1440

### One thing for the author to decide

The first screenshot shows it plainly: a signed-out panel now carries **two** sign-in affordances, `Sign In` at the top and `Sign in` in the account slot. That is the direct consequence of the requested change, and I do not think it reintroduces this PR's bug — the original defect was a *contradiction* (Sign In above a signed-in user's chip), whereas two controls that agree are redundant, not contradictory. Reasoning and a two-line alternative are in [this thread](https://github.com/recoupable/chat/pull/1914#discussion_r3685042487).

### Review findings

Validated against the code before acting, not taken on trust.

| Finding | Verdict | Action |
|---|---|---|
| @sweetmantech — return a sign-in button instead of `null` | **Valid** | new `SignInSlotButton` in the account slot, wired to Privy `login`; 2 tests |
| codex P2 / cubic P2 — wallet-only users treated as signed out | **Valid, and a regression I introduced** | see below |
| cubic P2 — stale `userData` can render the previous user's profile after an identity change | **Pre-existing, not from this PR** | the old code was `if (!userData) return skeleton`, so lingering data rendered the same way before. Genuinely worth fixing, but it needs `useUser` to associate `userData` with the current Privy identity, which is a bigger change than this PR's scope. Not addressed here |

**The regression, in full:** gating on Privy's `authenticated` flag alone would have hidden the profile chip entirely for Farcaster mini-app visitors. [`useAutoLogin`](https://github.com/recoupable/chat/blob/main/hooks/useAutoLogin.ts) deliberately skips Privy login when `isMiniApp`, while [`useUser`](https://github.com/recoupable/chat/blob/main/hooks/useUser.tsx) still derives `address` from `wagmiAddress` and bootstraps the account from it — so those users hold a real `userData` with `authenticated === false`. `getUserProfileState` now takes `hasWallet` and only reads `signed-out` when **neither** a Privy session **nor** a wallet is present. Two tests cover the wallet-only ready and loading paths.

Also removed a pre-existing unused `secondaryName` in the same file so it lints clean.

Suite **294 passed / 78 files**, `tsc --noEmit` clean, `eslint` clean on the touched files.

### Note unrelated to this PR

The signed-in screenshot shows **"Recoupable Pro: Active"** on `sweetman@recoupable.com`, a brand-new account that lands on `/setup` with an empty roster. That matches the finding recorded under *Findings outside this issue's scope* in [chat#1912](https://github.com/recoupable/chat/issues/1912) — new accounts appear to be seeded on the Pro allotment. Not caused by and not addressed in this PR.

PR #1915, opened 2026-07-30
fix(catalog): the report tab stops dead-ending cross-account and anonymous (chat#1912 row 1)

sweetmantech:
Closes row 1 of [chat#1912](https://github.com/recoupable/chat/issues/1912).

## Why

Measurements are **account-scoped**; the catalog is **not**. So `GET /api/catalogs/{id}/measurements` returns 404 in two very different situations, and the report tab collapsed both into one message:

> No valuation found for this catalog. This catalog has no measured play counts linked to your account yet. **Run a valuation from recoupable.dev**, or add songs in the Manage songs tab and ask Recoup to measure them.

That is wrong in both directions:

- **For the owner mid-measurement** it reads as "nothing here", so the customer re-runs the whole valuation. Reproduced on prod 2026-07-30; it is exactly what a live customer did on 07-27 (two `playcount_snapshots` runs, 209s apart).
- **For a different account** it claims no valuation exists when one does, and the only next step throws a signed-in customer back to the marketing site at the moment of peak interest.

A third state was wrong too: **anonymous** visitors got *"Something went wrong loading the measurements"*, because `useCatalogMeasurements` is gated on `authenticated` and the query never runs. A disabled query is not a failure.

## What changed

- **`lib/catalog/getCatalogReportState.ts`** — pure decision returning `loading | signed-out | measuring | other-account | error | ready`. The auth check deliberately precedes the error checks, since a disabled query leaves neither data nor error behind.
- **`lib/catalog/getCatalogReportEmptyCopy.ts`** — copy per state, kept out of the component so the wording is testable.
- **`hooks/useOwnsCatalog.ts`** — resolves ownership from the viewer's own catalog list, which is what separates "measuring" from "other-account". Exposes `isResolved` so a still-loading list is never mistaken for non-ownership.
- **`CatalogReportEmptyState`** replaces `CatalogReportError` (deleted, no other callers).
- **While measuring, the report polls every 5s** and resolves on its own. This is the part that removes the reason to re-run.

New copy:

| State | Says |
|---|---|
| signed-out | "Sign in to see this report" + a Sign in button (opens Privy in place) |
| measuring | "Measuring your catalog" + spinner, and explicitly *"There is no need to run the valuation again."* |
| other-account | "This catalog was measured by another account" + "Go to your catalogs" → `/catalogs` |
| error | unchanged wording, still points at the Manage songs tab |

## Deliberately not in this PR

**The songs/report access mismatch.** `GET /api/catalogs/songs` returns **200 unauthenticated**, so the Manage songs tab serves a stranger all 24 tracks and ISRCs on the same page that withholds the valuation. The issue's Works-when step 4 covers it. Closing that is an **api** change and a product decision about whether catalogs are public, so it is not bundled into a chat-side copy-and-state fix. Row 1's `other-account` copy is written to be honest either way ("The songs are listed under Manage songs, but the play counts and valuation belong to the account that measured them"). Flagging explicitly so this does not merge as if the access model were settled.

## Tests

TDD red→green (both RED as module-not-found first):

- `lib/catalog/__tests__/getCatalogReportState.test.ts` — 8 cases, one per branch, including the anonymous no-error case and 500-vs-404.
- `lib/catalog/__tests__/getCatalogReportEmptyCopy.test.ts` — 16 cases, including a guard that **no state's copy or CTA contains `recoupable.dev`** (the architecture decision recorded on the issue) and a house-style em-dash guard.

Full suite **306 passed / 77 files**, `tsc --noEmit` clean.

## Verification

Preview verification against the issue's four-step **Works when** script will be posted as a PR comment once the preview builds. Note it needs the api `test` branch synced with `main` first, since chat previews resolve to `test-recoup-api`.

---
## Summary by cubic
Fixes the Catalog report tab so anonymous and cross‑account viewers no longer dead‑end, and owners mid‑measurement see clear progress with auto‑refresh. Refactors state and polling into composed hooks and addresses review feedback (chat#1912 row 1, chat#1915).

- **Bug Fixes**
  - Added `getCatalogReportState` to clearly handle: loading, signed‑out, measuring, other‑account, error, ready; signed‑out no longer reads as an error and shows “Sign in” immediately.
  - While measuring, auto‑polls every 5s until ready to avoid duplicate valuations.
  - Missing‑measurement check keys off HTTP 404 status prefix (not substrings); unresolved ownership renders an error instead of guessing “other‑account”.

- **Refactors**
  - Moved report data/state/polling into `useCatalogReport` and `useMeasuringPoll`, keeping `CatalogReportContent` presentational.
  - Introduced `useOwnsCatalog` with `ownershipUnknown`; moved empty‑state copy to `getCatalogReportEmptyCopy`.
  - Replaced `CatalogReportError` with `CatalogReportEmptyState`.
  - Added tests for state, copy, and ownership resolution.

<sup>Written for commit 0bb24e0ddca370ddf9843dd3fb08b694dac27edb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


components/Catalog/report/CatalogReportContent.tsx
```diff
@@ -33,20 +38,45 @@ const CatalogReportContent = ({ catalogId }: CatalogReportContentProps) => {
...
+  });
+
+  // The owner's measurement is still running, so pull again instead of leaving
+  // them on an empty report to re-run the whole valuation (chat#1912 row 1).
+  const { refetch } = measurementsQuery;
+  useEffect(() => {
+    if (state !== "measuring") return;
+    const id = setInterval(() => refetch(), MEASURING_POLL_MS);
+    return () => clearInterval(id);
+  }, [state, refetch]);
```
sweetmantech: OCP 
- actual: net new hooks added to existing component
- required: put useEffects in a hook like useOwnsCatalog.

sweetmantech: Fixed in `0bb24e0d`.

The component no longer owns any of it. Composition instead:

- **`hooks/useCatalogReport.ts`** composes the measurements, songs, auth and ownership reads and derives the single report state.
- **`hooks/useMeasuringPoll.ts`** owns the while-measuring interval, so the `useEffect` lives in a hook rather than in the component.
- **`CatalogReportContent`** is presentational again: it calls one hook and renders what the state describes. Its only remaining hook is the pre-existing `useMemo` for the release rollups.

Net effect on the component: `usePrivy`, `useCatalogMeasurements`, `useCatalogReportSongs`, `useOwnsCatalog`, `getCatalogReportState` and the polling effect all left it, and one `useCatalogReport` call replaced them.

sweetmantech:

Preview **https://chat-6eev0xop0-recoup.vercel.app**, confirmed built from head `0bb24e0d` (deployment looked up by sha, not by branch).

Fixtures from [chat#1912](https://github.com/recoupable/chat/issues/1912): **catalog A** = `41aa8ad9-a347-438f-b2b4-998f49f8d4ec` (Ben's, measured by another account), **catalog B** = `006ca598-0ee3-480a-9e78-55bbe6354ff2` (the viewer's own), **account B** = `10c76dc5-360f-4678-86a1-efeda71d4f2e`.

### Results

| # | Check | Documented / before | Actual on preview | |
|---|---|---|---|---|
| 1 | Signed in as another account, open catalog A | "No valuation found for this catalog … Run a valuation from **recoupable.dev**" | **"This catalog was measured by another account"**, CTA `Go to your catalogs` → `/catalogs` | ✅ |
| 2 | Anonymous, open catalog A | "Couldn't load this report / Something went wrong loading the measurements" | **"Sign in to see this report"** + Sign in button that opens Privy **in place** | ✅ |
| 3 | Measuring state | progress within 1s, resolves without a re-run | **not reproduced** — see below | ⚠️ |
| 4 | Songs and report agree on access | both visible or both withheld | `GET /api/catalogs/songs` **200 unauthenticated**, `GET /…/measurements` **401** | ❌ known gap, out of scope |
| 5 | Owner regression control (catalog B) | full report | unchanged: **$1.6K** band, 160K lifetime streams, **24 of 24** tracks measured, 24 releases | ✅ |
| 6 | No off-app CTA anywhere | n/a | asserted programmatically: **zero** occurrences of `recoupable.dev` in page text or in any `href`, on both the anonymous and cross-account states | ✅ |
| 7 | Signed-out gets the CTA without waiting on songs (review P3) | skeleton until songs resolve | CTA renders immediately, no skeleton | ✅ |

### Screenshots

**Anonymous on catalog A** — the state that used to read "Something went wrong":

anonymous sign-in state

**Signed in as account B on catalog A** — the headline fix:

cross-account state

**Account B's own catalog** — regression control, report renders exactly as before:

owner report control

### Row 3: the measuring state was not reproduced, and the acceptance criterion was wrong

I could not reach it on the preview, and checking why turned up a mistake in how [chat#1912](https://github.com/recoupable/chat/issues/1912) specified this.

The Works-when step assumed a window where a freshly seeded catalog exists but has not been measured yet. **That window does not exist on the seeding path.** `runValuationHandler` measures first and calls `createSnapshotCatalog` afterwards, so the catalog row is created *after* its measurements. Measured on the recreation account: this run's `song_measurements` landed at `14:30:54.905`, the `catalogs` row at `14:30:55.644` — measurements **739ms earlier**. There is no point at which you can open a just-seeded catalog and find it unmeasured.

So the state is real but reachable by a different route: a catalog that has songs and no measurements *for this viewer* — songs added through Manage songs or a CSV upload, then measured later. That is worth verifying before the row is closed, and it needs a fixture I did not want to fabricate on prod data.

What is verified: the state machine that produces it is unit-tested (`getCatalogReportState`, 11 cases including this branch), and the three sibling states it shares that machine with all behave correctly live, which exercises the same code path.

**Recommendation:** keep row 1 open for the measuring case, and re-word the issue's Works-when step 3 to the CSV/Manage-songs route. I have not edited the issue yet — say the word and I will.

### Row 4 is a real gap, deliberately not closed here

`GET /api/catalogs/songs?catalog_id=41aa8ad9-…` returns **HTTP 200 with no credentials**, so the Manage songs tab still serves a stranger all 24 of Ben's tracks and ISRCs on the same page that now correctly withholds the valuation. Closing that is an api change plus a product decision about whether catalogs are public, which is why this PR's `other-account` copy is written to be true either way. Flagging it so this does not merge as if the access model were settled.

### Review findings

All four validated against the code before acting, none rubber-stamped.

| Finding | Verdict | Action |
|---|---|---|
| @sweetmantech — OCP: net new hooks added to the component | **Valid** | `useCatalogReport` + `useMeasuringPoll` now own composition and the effect; component is presentational again |
| cubic P1 / codex P2 — disabled catalog query treated as a resolved empty list | **Valid, real bug** | `useOwnsCatalog` derives resolution from `isSuccess`/`isError`. `useCatalogs` is `enabled: !!accountId && authenticated`, and a disabled TanStack v5 query reports `isPending` true / `isFetching` false, so `!isLoading` was already true while the account resolved — a signed-in owner would have been told their own catalog belonged to someone else. The same trap is documented in `SetupValuation.tsx`, which is what confirmed it |
| cubic P2 — `404` matched anywhere in the error text | **Valid** | matches the `HTTP 404` status prefix now, so a 500 whose body mentions 404 stays an error |
| cubic P3 — signed-out visitors wait on the songs read | **Valid** | signed-out visitors no longer gate on `songsQuery`; verified live as row 7 above |

Also added, not from a review: a failed catalog list now reports `ownershipUnknown` and renders the error state instead of guessing "other-account".

Suite **314 passed / 78 files**, `tsc --noEmit` and `eslint` clean.

PR #1913, opened 2026-07-30
feat(nav): link Catalogs from the side menu (chat#1912 row 3)

sweetmantech:
Closes row 3 of [chat#1912](https://github.com/recoupable/chat/issues/1912).

## Why

The app's only navigation lists exactly *Agents / Tasks / Files*. **Catalogs** — the valuation payoff surface, and the conversion target of the whole funnel — is unreachable from the menu; the catalog page is enterable only via direct URL or its own back-link.

Reported by a live outreach customer on 2026-07-29 (account `148e1644`):

> I guess I'm waiting on the newsletter because while I hear what Recoup I'm not sure where to find the tooling in the site.

Confirmed on prod 2026-07-30 at both breakpoints — 390px showed only New Chat / Agents / Tasks / Files, and the 1440px rail was four unlabeled icons. Screenshots in the issue.

## What changed

- New `components/Sidebar/CatalogsNavItem.tsx`, mirroring `FilesNavItem` exactly (same `NavButton`, `book` icon from the existing resolver, `aria-label="View catalogs"`). No new dependency, no new icon asset.
- `SecondaryNav` takes `isCatalogs` and renders the item first in the tool list — Catalogs is the payoff surface, so it leads.
- Both call sites wired with the pathname-derived active state the siblings already use: `Menu` (desktop rail) and `SideMenu` (mobile).

Catalogs leads the list rather than being appended, because the whole point of the report is that a customer arrives wanting the number.

## Tests

`components/Sidebar/__tests__/SecondaryNav.test.tsx` — new, TDD red→green:

1. Catalogs is present in the nav (RED before: `Unable to find an accessible element with the role "button" and name /view catalogs/i`)
2. clicking it navigates to `catalogs` (same RED)
3. regression guard: Agents / Tasks / Files still render (GREEN throughout, so the addition is proven additive)

Full suite **285 passed / 76 files**, `tsc --noEmit` clean.

## Verification

Preview verification against the issue's **Works when** script (390px + 1440px, navigational only, no pasted URLs) will be posted as a PR comment once the preview builds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Adds Catalogs to the sidebar and mobile side menu, placed first in the tools list so users can open catalogs directly from navigation. Addresses chat#1912 row 3 by making catalogs discoverable.

- **New Features**
  - Added Catalogs to desktop rail and mobile menu using `NavButton` with the "disc" icon and aria-label "View catalogs".
[...]

PR #1916, opened 2026-07-30
fix(hydration): stop the mobile media query running during hydration (chat#1912 row 5)

sweetmantech:
Closes row 5 of [chat#1912](https://github.com/recoupable/chat/issues/1912).

## Why

`Uncaught Error: Minified React error #418` fires on **every cold load**, signed out and signed in, on `/` and `/catalogs/{id}`. React regenerates the entire tree on the client each time, so the whole app pays a full re-render on first paint.

The error's args are `["HTML", ""]`, which in React 19's [`throwOnHydrationMismatch`](https://github.com/facebook/react/blob/main/packages/react-dom-bindings/src/client/ReactFiberConfigDOM.js) is `fromText ? "text" : "HTML"` — an **element** mismatch, not a text one. That ruled out the usual suspects (`Date` formatting, locale, `Math.random()`), and SSR output was confirmed byte-identical across fetches, so the server was not the variable side.

Reproducing against a **dev build** made React name the node:

```
<Header>
  <div className="z-[50] fix...">
    <button>
+   <button
+     type="button"
+     onClick={function handleAddArtist}
+     aria-label="Add a new artist"
+   >
```

The `+` is client-only output. [`Header.tsx`](https://github.com/recoupable/chat/blob/main/components/Header/Header.tsx#L52) renders that button behind `isMobile && !isArtistSelected`, and `useIsMobile` wraps usehooks-ts `useMediaQuery`, whose `initializeWithValue` **defaults to `true`**:

```js
const [matches, setMatches] = useState(() => {
  if (initializeWithValue) return getMatches(query); // window.matchMedia, during the first render
  return defaultValue;
});
```

On the server `getMatches` returns `defaultValue` (`false`); on the client the *hydration* render already evaluates `matchMedia`, so at any viewport under 768px the two renders disagree and an extra `<button>` appears. Nothing about this is mobile-only in consequence: the mismatch regenerates the whole tree for that visitor.

## What changed

One option on `hooks/useIsMobile.tsx`: `initializeWithValue: false`. The first render now matches the server, and the real viewport value arrives via the hook's existing layout effect — so the mobile UI still appears immediately, it just no longer contradicts the server's HTML.

`components/Sidebar/RecentChats/useRecentChats.ts` uses a different hook (`useMobileDetection`) which is already effect-based and correct, so it needed no change.

## Verification

Reproduced and fixed against a local dev build at a 500px viewport:

| | Before | After |
|---|---|---|
| `/` cold load | `Hydration failed because the server rendered HTML didn't match the client` + the `<Header>` diff above | **no hydration error** |
| `/catalogs` cold load | same | **no hydration error** |

The only console error remaining on localhost is Privy's `frame-ancestors` CSP rejection, which is expected off-domain and unrelated.

## Tests

`hooks/__tests__/useIsMobile.test.tsx` — 3 cases, TDD red→green. The first two were RED against the real defect (the hook called `useMediaQuery` with no options object). They assert the hook does not evaluate the query during the hydration render, keeps the 768px breakpoint, and returns what the query reports.

Full suite **285 passed / 76 files**, `tsc --noEmit` clean.

## Verification on preview

The issue's **Works when** script (cold, signed-out loads of `/`, `/catalogs`, `/catalogs/{id}`, console clean of #418 / #423 / #425) will be re-run against the preview and posted as a PR comment.

---
## Summary by cubic
Prevent the mobile media query from running during hydration to stop React hydration mismatches (#418). Fixes chat#1912 row 5 and removes the full re-render on cold loads of `/` and `/catalogs`.

- **Bug Fixes**
  - Use `useMediaQuery("(max-width: 768px)", { initializeWithValue: false })` from `usehooks-ts` in `useIsMobile` so the first client render matches SSR and no extra Header button appears on narrow viewports.
  - Add tests for `useIsMobile` to assert no query during hydration, the 768px breakpoint, and that the return mirrors the query.

<sup>Written for commit c702c1034f89f5f3d9ba9c088bb785edb2642b69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mobile and desktop layout detection during initial page loading.
  * Prevented hydration mismatches and visual differences between server-rendered and client-rendered content.

### DIFF
--- a/hooks/__tests__/useIsMobile.test.tsx
+++ b/hooks/__tests__/useIsMobile.test.tsx
@@ -1,0 +1,44 @@
+// @vitest-environment jsdom
+import { renderHook } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import useIsMobile from "@/hooks/useIsMobile";
+
+const useMediaQuery = vi.hoisted(() => vi.fn(() => false));
+
+vi.mock("usehooks-ts", () => ({ useMediaQuery }));
+
+describe("useIsMobile", () => {
+  /**
+   * chat#1912 row 5. usehooks-ts defaults `initializeWithValue` to true, which
+   * evaluates window.matchMedia during the very first render — including the
+   * hydration render. The server has no matchMedia and renders the desktop
+   * branch, so on a narrow viewport the client's first render disagreed and
+   * Header emitted an extra "Add Your Artist" button the server never sent.
+   * React threw #418 and regenerated the whole tree on every page load.
+   */
+  it("does not evaluate the media query during the hydration render", () => {
+    renderHook(() => useIsMobile());
+
+    expect(useMediaQuery).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.objectContaining({ initializeWithValue: false }),
+    );
+  });
+
+  it("keeps the 768px breakpoint", () => {
+    renderHook(() => useIsMobile());
+
+    expect(useMediaQuery).toHaveBeenCalledWith(
+      "(max-width: 768px)",
+      expect.anything(),
+    );
+  });
+
+  it("returns what the media query reports", () => {
+    useMediaQuery.mockReturnValue(true);
+
+    const { result } = renderHook(() => useIsMobile());
+
+    expect(result.current).toBe(true);
+  });
+});
--- a/hooks/useIsMobile.tsx
+++ b/hooks/useIsMobile.tsx
@@ -1,7 +1,21 @@
 import { useMediaQuery } from "usehooks-ts";
 
+/**
+ * Viewport check for the mobile breakpoint.
+ *
+ * `initializeWithValue: false` is required, not cosmetic. usehooks-ts defaults
+ * it to true, which evaluates `window.matchMedia` during the first render —
+ * including the hydration render. The server has no matchMedia and renders the
+ * desktop branch, so on a narrow viewport the client's first render disagreed
+ * and `Header` emitted an "Add Your Artist" button the server never sent.
+ * React treated that as a hydration mismatch (#418) and regenerated the entire
+ * tree on every page load (chat#1912 row 5). With this off the first render
+ * matches the server, and the real value arrives in an effect.
+ */
 const useIsMobile = () => {
-  const isMobile = useMediaQuery("(max-width: 768px)");
+  const isMobile = useMediaQuery("(max-width: 768px)", {
+    initializeWithValue: false,
+  });
 
   return isMobile;
 };


sweetmantech:

Preview **https://chat-k92m5496q-recoup.vercel.app**, built from head `c702c103` (deployment resolved by sha).

**Rebased onto `main` first**, so the preview includes the three rows already merged today ([#1915](https://github.com/recoupable/chat/pull/1915), [#1914](https://github.com/recoupable/chat/pull/1914), [#1913](https://github.com/recoupable/chat/pull/1913)). Verifying the pre-rebase branch would have tested a tree that no longer matches what merges. The rebase is confirmed live in the results below: the menu shows both `View catalogs` (#1913) and the `Sign in` slot (#1914).

**The viewport was set before each load, never after.** This bug only exists below 768px — above it the server and client agree and nothing reproduces — and resizing after hydration proves nothing, because hydration has already happened.

### Results

| # | Check | Before | Actual | |
|---|---|---|---|---|
| 0 | **Control:** prod `/` cold, signed out, 500px wide | — | `Uncaught Error: Minified React error #418` | 🔴 reproduces |
| 1 | Preview `/` cold, signed out, same viewport | #418 on every load | **no console output at all** | ✅ |
| 2 | Preview `/catalogs` cold | #418 | **none** | ✅ |
| 3 | Preview `/catalogs/{id}` cold | #418 | **none** | ✅ |
| 4 | Preview `/` cold at **1440px** | — | **none** | ✅ |
| 5 | **Regression control:** the mobile UI the mismatch came from | header renders `Add Your Artist` under 768px | still renders (`Add a new artist` present at 500px) | ✅ |
| 6 | Rebase actually applied | — | menu shows `View catalogs` and the `Sign in` slot | ✅ |

Row 0 is the point of the run: the same browser, same viewport, same signed-out state throws #418 against production and is silent against this preview. Row 5 is the one that could have gone wrong — the fix defers the media-query result to an effect, so the risk was that the mobile UI never appears at all. It appears.

### What was actually wrong

usehooks-ts defaults `useMediaQuery`'s `initializeWithValue` to **true**, which evaluates `window.matchMedia` during the *first* render — including the hydration render:

```js
const [matches, setMatches] = useState(() => {
  if (initializeWithValue) return getMatches(query); // runs during hydration
  return defaultValue;
});
```

The server has no `matchMedia` and takes the desktop branch, so below 768px the client's first render disagreed and [`Header`](https://github.com/recoupable/chat/blob/main/components/Header/Header.tsx#L52) emitted an extra `<button aria-label="Add a new artist">` the server never sent. React treated that as a mismatch and **regenerated the entire tree on every page load**, for signed-out and signed-in visitors alike.

Pinning it took three narrowing steps, recorded because the first two are reusable:

1. The error's args are `["HTML", ""]`. In React 19's `throwOnHydrationMismatch` the first arg is `fromText ? "text" : "HTML"`, so this was an **element** mismatch — which ruled out the usual suspects (date formatting, locale, `Math.random()` in text).
2. SSR output was **byte-identical across two fetches** (65,425 bytes), so the server was not the variable side.
3. A local dev build then made React name the offending node via `describeDiff`.

`components/Sidebar/RecentChats/useRecentChats.ts` uses a different hook (`useMobileDetection`) which is already effect-based and correct, so it needed no change.

Suite **332 passed / 82 files** after the rebase, `tsc --noEmit` clean.

### No review findings

No bot or human comments were left on this PR, so there was nothing to triage.

PR #1917, opened 2026-07-30
fix(catalog): cross-account copy stops promising the songs tab (chat#1912 row 6)

sweetmantech:
Third and final link of row 6's chain in [chat#1912](https://github.com/recoupable/chat/issues/1912).

**Merge order: [docs#282](https://github.com/recoupable/docs/pull/282) → [api#802](https://github.com/recoupable/api/pull/802) → this.** It is safe to merge at any point (the new wording is true before and after), but it only becomes *necessary* once api#802 lands.

## Why

[chat#1915](https://github.com/recoupable/chat/pull/1915) shipped this copy for a cross-account viewer:

> This catalog was measured by another account.
> **The songs are listed under Manage songs**, but the play counts and valuation belong to the account that measured them.

That sentence was written to be true either way while the access model was undecided — at the time, `GET /api/catalogs/songs` really did return 200 to anyone. [api#802](https://github.com/recoupable/api/pull/802) makes it **403** for a non-owner, so the copy would be pointing a stranger at a tab that now fails.

## What changed

```diff
- The songs are listed under Manage songs, but the play counts and valuation
- belong to the account that measured them.
+ Its play counts and valuation belong to the account that measured it. Run a
+ valuation on your own catalog to see this report.
```

Same title, same `Go to your catalogs` CTA. The replacement states only what remains true and offers a next step instead of a dead reference.

## Tests

TDD red→green: a new case asserts the `other-account` body does not mention "Manage songs", so this cannot silently regress if the copy is edited again later. The existing guards in that file still pass unchanged — no `recoupable.dev` in any state's copy or CTA, and no em dashes.

Full suite **333 passed / 82 files**, `tsc --noEmit` and `eslint` clean.

## Verification

Nothing to verify on a preview until api#802 is merged — before that, the songs tab still returns 200 and both wordings are accurate. Row 6's Works-when step 4 ("re-read the other-account wording and confirm it does not describe behaviour the API no longer has") covers this once the chain is complete.

---
## Summary by cubic
Rewords the cross-account catalog report empty state so it no longer points to “Manage songs”. Adds an adaptive CTA that keeps viewers in-app with a clear next step.

- **Bug Fixes**
  - Updated body to: "Its play counts and valuation belong to the account that measured it. Here is how to get this report for your own music." CTA adapts: has catalogs → "Go to your catalogs" (`/catalogs`); none → "Value your catalog" (`/setup/artists`).
  - Added `hasOwnCatalogs` to `useOwnsCatalog` → `useCatalogReport` → `CatalogReportEmptyState`, and tests to guard wording, ensure cross-account always has a CTA, and keep measuring with no CTA.

<sup>Written for commit e781cd6fc89e8812163316e674a882e33fc9cdd1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the report empty state for another-account views, tailoring the call-to-action based on whether the viewer has their own catalogs.
  * Updated guidance so users are directed to the correct valuation flow to view a report for their own music when applicable.
  * Clarified in the empty-state messaging how play counts and valuation relate to the measuring account.

### DIFF
--- a/lib/catalog/getCatalogReportEmptyCopy.ts
+++ b/lib/catalog/getCatalogReportEmptyCopy.ts
@@ -12,6 +12,11 @@ export interface CatalogReportEmptyCopy {
   cta?: { label: string; href?: string; action?: "login" };
 }
 
+export interface CatalogReportViewer {
+  /** Whether the signed-in viewer has any catalog of their own. */
+  hasOwnCatalogs: boolean;
+}
+
 const COPY: Record<CatalogReportEmptyState, CatalogReportEmptyCopy> = {
   "signed-out": {
     title: "Sign in to see this report",
@@ -24,7 +29,7 @@ const COPY: Record<CatalogReportEmptyState, CatalogReportEmptyCopy> = {
   },
   "other-account": {
     title: "This catalog was measured by another account",
-    body: "Its play counts and valuation belong to the account that measured it. Run a valuation on your own catalog to see this report.",
+    body: "Its play counts and valuation belong to the account that measured it. Here is how to get this report for your own music.",
     cta: { label: "Go to your catalogs", href: "/catalogs" },
   },
   error: {
@@ -40,6 +45,20 @@ const COPY: Record<CatalogReportEmptyState, CatalogReportEmptyCopy> = {
  */
 export function getCatalogReportEmptyCopy(
   state: CatalogReportEmptyState,
+  viewer: CatalogReportViewer,
 ): CatalogReportEmptyCopy {
-  return COPY[state];
+  const copy = COPY[state];
+
+  // A viewer who cannot see this catalog still needs somewhere to go, and
+  // "your catalogs" is only somewhere for people who have one. A stranger
+  // following a shared link usually has none, so point them at the step that
+  // creates one instead of an empty page.
+  if (state === "other-account" && !viewer.hasOwnCatalogs) {
+    return {
+      ...copy,
+      cta: { label: "Value your catalog", href: "/setup/artists" },
+    };
+  }
+
+  return copy;
 }
--- a/lib/catalog/getCatalogReportEmptyCopy.ts
+++ b/lib/catalog/getCatalogReportEmptyCopy.ts
@@ -24,7 +24,7 @@ const COPY: Record<CatalogReportEmptyState, CatalogReportEmptyCopy> = {
   },
   "other-account": {
     title: "This catalog was measured by another account",
-    body: "The songs are listed under Manage songs, but the play counts and valuation belong to the account that measured them.",
+    body: "Its play counts and valuation belong to the account that measured it. Run a valuation on your own catalog to see this report.",
     cta: { label: "Go to your catalogs", href: "/catalogs" },
   },
   error: {
--- a/components/Catalog/report/CatalogReportContent.tsx
+++ b/components/Catalog/report/CatalogReportContent.tsx
@@ -27,7 +27,7 @@ interface CatalogReportContentProps {
  * this component only renders what that state describes.
  */
 const CatalogReportContent = ({ catalogId }: CatalogReportContentProps) => {
-  const { state, measurements, songs, totalSongs } =
+  const { state, measurements, songs, totalSongs, hasOwnCatalogs } =
     useCatalogReport(catalogId);
 
   const releases = useMemo(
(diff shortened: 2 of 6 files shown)


lib/catalog/getCatalogReportEmptyCopy.ts
sweetmantech, line 1: Joy - How can we put an action button on the screen to help create a happy path rather than leave this as a dead end with text?

sweetmantech: Good catch, and it was worse than "just text" — there **was** a button, but it was only a happy path for people who already have catalogs. `Go to your catalogs` sends a stranger who followed a shared link to an **empty page**: a second dead end, one click further in. That is the exact shape of the original bug.

Fixed in `e781cd6f` by making the action adapt to the viewer:

| Viewer | Action |
|---|---|
| has catalogs | **Go to your catalogs** → `/catalogs` |
| has none (the shared-link stranger) | **Value your catalog** → `/setup/artists` |

`/setup/artists` is the step that actually creates a catalog, so the person who arrived curious about someone else's number is now one click from getting their own. It stays in-app, which keeps the architecture decision on this issue intact (the empty state never bounces anyone to recoupable.dev).

Body copy now leads into the action rather than explaining the refusal twice: *"Its play counts and valuation belong to the account that measured it. Here is how to get this report for your own music."*

`useOwnsCatalog` already loads the viewer's catalog list to decide ownership, so the flag costs no extra request.

Four tests cover it, including one asserting the cross-account state **always** has something to click, in either branch — so this cannot silently regress to a dead end again.

sweetmantech:

Preview **https://chat-7tdt85jib-recoup.vercel.app**, built from head `e781cd6f` (resolved by sha), signed in as `sweetman@recoupable.com` — chosen deliberately because that account has **no catalogs**, which is the branch the review comment created.

**The api `test` branch was synced with `main` first** (it was 2 commits behind — [api#800](https://github.com/recoupable/api/pull/800) and [api#802](https://github.com/recoupable/api/pull/802)). Chat previews resolve to `test-recoup-api`, so without that sync this preview would have tested against an API that still served the songs tab, and the whole point of this PR would have been unverifiable.

### Results

| # | Check | Actual | |
|---|---|---|---|
| 1 | Non-owner opens a shared catalog URL | **"This catalog was measured by another account"** | ✅ |
| 2 | Body no longer promises the songs tab | `"listed under Manage songs"` **absent** from the page | ✅ |
| 3 | Adaptive CTA for a viewer with no catalogs | **"Value your catalog"** → `/setup/artists` | ✅ |
| 4 | The CTA actually goes somewhere useful | click → lands on **`/setup/artists`**, "Step 1 of 2 — Confirm your roster" | ✅ |
| 5 | **Manage songs tab for a non-owner** | renders the same honest state; **no tracklist, no ISRCs, no "Found 24 songs"** | ✅ |
| 6 | Still no off-app CTA | zero `recoupable.dev` in text or any `href` | ✅ |

**Row 5 is the reason this PR exists.** With api#802 live on `test`, that tab now refuses a non-owner. The old copy pointed them straight at it. Had this shipped without the reword, the page would have said *"The songs are listed under Manage songs"* directly above a tab that returns 403 — the same self-contradiction the issue was opened for, just relocated.

**Row 4 matters more than it looks.** The old CTA (`Go to your catalogs`) was technically a button, but for a stranger following a shared link it led to an **empty page** — a dead end one click deeper. Now it leads to the step that actually creates a catalog.

### Screenshots

**Non-owner on a shared catalog — honest state with a real next action:**

cross-account with Value your catalog

**Manage songs tab, same viewer — the tracklist and ISRCs are gone:**

manage songs no longer leaks

**The CTA lands on the roster step, not an empty page:**

CTA lands on setup

### Review

@sweetmantech asked for a happy path rather than a dead end. Addressed in `e781cd6f` and replied on [the thread](https://github.com/recoupable/chat/pull/1917#discussion_r3686402094): the action adapts, `/catalogs` for viewers who have catalogs and `/setup/artists` for those who do not. One test asserts the cross-account state **always** has something to click in either branch, so it cannot regress to a dead end again.

Suite **337 passed / 82 files**, `tsc --noEmit` and `eslint` clean.

### Note

This is the last link of the row 6 chain ([docs#282](https://github.com/recoupable/docs/pull/282) → [api#802](https://github.com/recoupable/api/pull/802) → this). Both earlier links are merged, so **production currently refuses the songs tab while still telling non-owners to go there.** This should merge promptly rather than sit.

PR #1919, opened 2026-07-31
fix(onboarding): /setup/valuation polls its way out of measuring (chat#1912 row 9)

sweetmantech:
Closes row 9 of [chat#1912](https://github.com/recoupable/chat/issues/1912).

## Why

`/setup/valuation` renders `MeasuringCatalogPanel` — *"Measuring your catalog … this usually takes a minute"* — and then **never re-checks**.

`useCatalogMeasurements` sets `staleTime` and `refetchOnWindowFocus: false` and has **no `refetchInterval`**. The only poller in the codebase lives in `useCatalogReport`, which [chat#1915](https://github.com/recoupable/chat/pull/1915) built for `/catalogs/{id}`. So a signup who lands here while their seeded catalog is still measuring sits on static text until they refresh by hand.

That is the same dead end row 1 fixed, on a route row 1 did not cover — and this one is on the path the welcome email points at.

## What changed

- **`lib/onboarding/getSetupValuationStatus.ts`** — pure `loading | redirect | measuring | ready`. `catalogsPending` is checked first, because a pending list is not an empty one: `useCatalogs` is `enabled: !!accountId && authenticated`, and a disabled TanStack v5 query reports `isPending` true / `isFetching` false. Reading that as "no catalog" would redirect a signup away mid-load.
- **`hooks/useSetupValuation.ts`** — composes the catalog and valuation reads, owns the redirect, and **reuses the existing `useMeasuringPoll`** rather than introducing a second poller.
- **`SetupValuation`** renders what the status describes; its `useEffect` moved into the hook, per the OCP note on [chat#1915](https://github.com/recoupable/chat/pull/1915#discussion_r3684831744).

## Tests

- `getSetupValuationStatus` — 6 cases including the one that matters, "prefers loading over redirect when both could apply", which is the trap the original comment in this file warned about.
- `useMeasuringPoll` — **4 new cases**, the first direct coverage this hook has had since row 1 introduced it: it polls while measuring, does not when idle, stops when measuring ends, and stops on unmount. The last two matter because a poll that outlives its state keeps hitting the api for as long as the tab is open.
- The 5 existing `SetupValuation` tests pass unchanged through the new composition. They needed one added mock: the component now reaches `useQueryClient` through the hook, and those tests render it without a provider.

Full suite **347 passed / 83 files**, `tsc --noEmit` and `eslint` clean.

## Verification

Preview verification against row 9's Works-when — add a first artist on a fresh account, go straight to `/setup/valuation`, and **do not touch the page**; it should resolve to the baseline valuation on its own, and polling should stop once it does — will be posted as a PR comment.

---
## Summary by cubic
Fixes onboarding `/setup/valuation` to poll catalogs and measurements and handle loading states correctly, so the page exits “Measuring your catalog” on its own and avoids redirects during seeding or while the roster is still loading.

- **Bug Fixes**
  - Reuses `useMeasuringPoll`; stops when measuring ends or on unmount.
  - Treats pending `useCatalogs` and the artist roster as loading, not “no catalog”/“no artists” (`@tanstack/react-query` v5).
  - During seeding, shows measuring if there are artists but no catalog yet; redirects only when no artists and no catalog or the list failed.
  - Invalidates `["catalogs"]` and `["catalog-measurements"]` while polling.

- **Refactors**
  - Added `getSetupValuationStatus` (loading | redirect | measuring | ready).
  - Added `useSetupValuation` to compose reads, own redirect, and drive polling.
  - `SetupValuation` now renders solely based on status.

<sup>Written for commit 1615d7fb24a0560dbbfc8465f8498fa58482e2fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved valuation setup flow with clearer loading, measuring, ready, and redirect states.
  * Automatically guides accounts without a catalog to the catalog setup process.
  * Continues tracking measurements while valuation setup is in progress.

* **Bug Fixes**
  * Improved handling of pending or failed catalog and valuation states.

### DIFF
--- a/components/Onboarding/SetupValuation.tsx
+++ b/components/Onboarding/SetupValuation.tsx
@@ -1,14 +1,11 @@
 "use client";
 
-import { useEffect } from "react";
 import Link from "next/link";
-import { useRouter } from "next/navigation";
 import { buttonVariants } from "@/components/ui/button";
 import { Skeleton } from "@/components/ui/skeleton";
 import ValuationHero from "@/components/Home/ValuationHero";
 import MeasuringCatalogPanel from "@/components/Onboarding/MeasuringCatalogPanel";
-import useCatalogs from "@/hooks/useCatalogs";
-import useHomeValuation from "@/hooks/useHomeValuation";
+import useSetupValuation from "@/hooks/useSetupValuation";
 import { cn } from "@/lib/utils";
 
 /**
@@ -17,27 +14,13 @@ import { cn } from "@/lib/utils";
  * to be a bare redirect to `/catalogs`, so the email's "See your baseline
  * valuation" link had no real destination.
  *
- * Falls back to `/catalogs` only when the account has no catalog to value, so a
- * cold-start signup still lands somewhere actionable rather than on an empty
- * hero.
+ * Status, the redirect for accounts with no catalog, and the measuring poll all
+ * live in `useSetupValuation`; this renders what the status describes.
  */
 const SetupValuation = () => {
-  const router = useRouter();
-  const valuation = useHomeValuation();
-  // `isPending`, not `isLoading`: useCatalogs is `enabled: !!accountId &&
-  // authenticated`, and a disabled TanStack Query v5 query reports
-  // isPending true / isFetching false — so isLoading is false while Privy is
-  // still resolving. Redirecting on that bounced every cold load of this
-  // route, which is exactly where the welcome email's payoff link points.
-  const { data, isPending, isError } = useCatalogs();
-  const hasCatalog = !!data?.catalogs?.length;
+  const { status, valuation } = useSetupValuation();
 
-  useEffect(() => {
-    if (isPending) return;
-    if (!hasCatalog || isError) router.replace("/catalogs");
-  }, [isPending, hasCatalog, isError, router]);
-
-  if (isPending) {
+  if (status === "loading" || status === "redirect") {
     return (
       <div className="mx-auto flex w-full max-w-xl flex-col gap-4 px-6 py-8">
         <Skeleton className="h-8 w-1/2 rounded-lg" />
@@ -47,8 +30,8 @@ const SetupValuation = () => {
   }
 
   // A catalog exists but has no valuation yet: routine while a seeded
-  // valuation is still measuring (chat#1889 row 8).
-  if (!valuation.show) return <MeasuringCatalogPanel />;
+  // valuation is still measuring (chat#1889 row 8). The hook polls it out.
+  if (status === "measuring" || !valuation.show) return <MeasuringCatalogPanel />;
 
   return (
     <section className="mx-auto flex w-full max-w-xl flex-col items-center gap-4 px-6 py-8">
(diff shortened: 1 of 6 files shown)


sweetmantech:
## Preview verification — 2026-07-31

Preview **https://chat-xk0csc8g2-recoup.vercel.app**, built from head `1615d7fb` (resolved by sha). Run on a **fresh signup**, because the seeding window exists exactly once per account.

### Results

| # | Check | Before | Actual | |
|---|---|---|---|---|
| 1 | Land on `/setup/valuation` right after adding the first artist | redirected to `/catalogs` — *"No catalogs found"* | **stays on `/setup/valuation`** | ✅ |
| 2 | Shows the measuring state while seeding runs | n/a (redirected away) | *"Measuring your catalog"* at **1s** | ✅ |
| 3 | Resolves **untouched** | never; required a manual refresh | hero renders **$1.6K, range $1.1K to $2.3K, 24 tracks measured** | ✅ |
| 4 | Polling stops once resolved | n/a | request counts identical after a further **20s** (`catalogs: 4→4`, `measurements: 1→1`) | ✅ |
| 5 | Never redirects during the wait | — | `redirected: false` across the whole observation | ✅ |

Timing for this account: artist added `01:26:37`, catalog created `01:26:59` — a **22-second** gap the route now sits through instead of bouncing out of.

resolved on route

### Two defects this found that the unit tests passed straight through

Both needed a real signup; the tests were green throughout because they supply already-resolved state.

**1. The fix was aimed at the wrong window.** The first run still redirected to an empty `/catalogs`. The database showed why: `createSnapshotCatalog` runs only *after* the measurements land, so the catalog appears **~15-22s after** the artist is added (measured twice: `00:59:04 → 00:59:19`, and `01:26:37 → 01:26:59`). For that entire window **there is no catalog at all**. The window this route was originally written for — catalog exists but is not yet measured — is about **one second**, because measurements are written first. So the original patch polled the narrow case and left the wide one redirecting away. An artist on the roster now means seeding is in flight, which is `measuring`, not `redirect`.

**2. A loading roster is not an empty one.** The second run *still* redirected. The roster loads asynchronously, so `sorted` is briefly empty on a fresh page load and read as "no artists". That is the same trap [the comment already in this file](https://github.com/recoupable/chat/blob/main/components/Onboarding/SetupValuation.tsx) warns about for the catalogs query — pending is not empty — which I then reproduced for the roster. `artistsPending` now holds the route in `loading` until both reads resolve.

### A measurement error worth recording

My first "has it resolved?" probe matched `/estimate goes stale|Set up your weekly report/`. **`Set up your weekly report` is also the CTA on the measuring panel**, so the probe reported "resolved" while the measuring state was still on screen. Re-measured against `estimate goes stale`, which appears only in the ready branch. A detector that matches both states of the thing you are testing will confirm whatever you hoped.

### Tests

`getSetupValuationStatus` 8 cases (including both traps above), `useMeasuringPoll` 4, `SetupValuation` 6. One pre-existing test was **removed** rather than adjusted: it asserted "no catalog → redirect", which this deliberately reverses; the two cases replacing it assert both halves — measuring when an artist exists, redirect when neither does.

Full suite **349 passed / 83 files**, `tsc --noEmit` and `eslint` clean.

### Residue

Three test accounts (`sweetman+row9*@recoupable.com`), one seeded catalog each. The first two were spent proving the defects above.

Issue #1923, opened 2026-08-03
✅ Chat SSE delivery: dropped streams no longer freeze the UI (fixed + prod-verified 2026-08-05)


sweetmantech:
> **✅ CLOSED — fixed and verified on prod 2026-08-05.** The original defect: a chat's SSE response ended at the ~120 s connection cap with a clean `data: [DONE]` and no `finish` chunk, so `useChat` stopped rendering while the workflow kept running — the UI froze with no error and no way back but a refresh.
>
> **Shipped, in merge order:** [api#809](https://github.com/recoupable/api/pull/809) resume route → [docs#286](https://github.com/recoupable/docs/pull/286) contract → [chat#1924](https://github.com/recoupable/chat/pull/1924) chunk counting → [chat#1934](https://github.com/recoupable/chat/pull/1934) SDK to current stable → [api#810](https://github.com/recoupable/api/pull/810) expose `x-workflow-run-id` to cross-origin JS → [api#814](https://github.com/recoupable/api/pull/814) send `finish` before closing on **every** exit path → [chat#1935](https://github.com/recoupable/chat/pull/1935) replace the hand-rolled recovery with `WorkflowChatTransport`.
>
> **Verified on prod** (`chat-cunwm87on-recoup` → `chat.recoupable.dev`, built from `main`): the exact prompt that used to freeze at two minutes ran **456 s (~7.6 min)**, streamed throughout, and settled on its own — composer back to its send control, email sent, `wrun_01KZ81CZJQHEYP3NPN8PR` completed, client and server in agreement at every check. Real prod data (Mac DeMarco at 9,069,137 Spotify followers), not fixtures.
>
> **Verified on preview beforehand:** 3/3 runs at **525 s / 983 s / 826 s**, every one 4–8× past the cap, each ending with client and workflow agreeing. Round 1 was 2/3 — its single failure was a `401 Authentication token expired` after the **laptop was shut mid-run**, which is tab suspension rather than a streaming defect; the retry written for it was deliberately rolled back.
>
[...]

PR #1924, opened 2026-08-03
fix(chat): reconnect a dropped response stream instead of freezing mid-turn

sweetmantech:
Client half of [chat#1923](https://github.com/recoupable/chat/issues/1923). Depends on [api#809](https://github.com/recoupable/api/pull/809) (the resume route) and [docs#286](https://github.com/recoupable/docs/pull/286) (the contract).

## Why

Reproduced on prod 2026-08-02 by replaying the reported prompt through the real UI. The `/api/chat` SSE response **ended at ~123 s** with a clean `data: [DONE]` and **no `finish` chunk**, while the workflow ran on to completion.

| | |
|---|---|
| UI froze at | **02:26:58** — exactly when the connection closed |
| Still frozen at | 02:36:12, nine minutes later |
| Run actually completed | 02:31:05 |
| Iterations delivered | **6 of 13** |
| Reconnect attempts made | **0** |

No error, no retry, no visible state change — the composer looked idle, as if the turn had finished. Only a page refresh recovered it.

## The SDK already had the machinery

`DefaultChatTransport.reconnectToStream` defaults to **`GET ${api}/${chatId}/stream`**. Our transport's `api` is `${baseUrl}/api/chat`, so the SDK has been prepared to reconnect to exactly the endpoint our docs describe — and which nothing ever implemented. We were missing both halves.

## What

| File | Role |
|---|---|
| `hooks/useVercelChat.ts` | `resume: true` — re-attach to an in-progress response on mount, so returning to a chat mid-turn keeps rendering. Wires `useStreamRecovery`. |
| `hooks/useStreamRecovery.ts` | Watches an in-flight turn for silence, calls `resumeStream()`. Also re-checks on `visibilitychange`. |
| `lib/chat/shouldRecoverStalledStream.ts` | The pure decision — unit-tested, no React. |
| `hooks/useChatTransport.ts` | `prepareReconnectToStreamRequest` attaches the bearer token. |

Composed-hook shape rather than growing `useVercelChat` internals, matching how the other chat hooks extend it.

## Design calls

**Silence-based, not duration-based.** A turn that is still streaming is healthy however long it runs; a turn that has gone quiet is suspect even if it just started. Keying on elapsed total time would fight the legitimate long turns api#808 just made possible.

**A cooldown, so a permanently dead stream isn't retried every tick.** Reconnects are cheap but not free, and a 204 ends the loop naturally once the run finishes.

**Auth on reconnect is not optional.** The resume route authenticates like every other endpoint; without `prepareReconnectToStreamRequest` the reconnect 401s and a dropped stream stays dropped. Easy to miss because the failure looks identical to having no recovery at all.

## Not included: precise `startIndex`

api#809 accepts `startIndex` for a gap-free resume, but `useChat` does not expose a received-chunk count, so there is no honest value to send. The SDK's default reconnect replays the stream and reconciles by message id, which is correct if slightly wasteful. Passing a *guessed* index would risk skipping content — worse than replaying it. Wiring an exact index is a follow-up once we track it properly; the route already supports it.

## Tests

`shouldRecoverStalledStream` — 7 cases, RED before GREEN: recovers on a stalled streaming turn; does not while chunks arrive; not once the turn is `ready`/`error`; recovers a `submitted` turn that never started; holds off while an attempt is in flight; honours the cooldown; treats a never-seen chunk timestamp as no reason to act.

- **Full chat suite: 363 tests passing.**
- **`tsc --noEmit`: 7 errors, identical to the `main` baseline (7), zero in any file this PR touches.**

## One thing I could not verify locally

`next build` in my worktree died with `WorkerError: Call retries were exceeded` — a worker crash, not a type or lint failure, which I attribute to the symlinked `node_modules` in the worktree rather than this diff. **CI's build is the authority here**; I'd rather flag that than claim a green build I didn't get. Preview verification against a real dropped stream is still owed and will be posted here.

---
## Summary by cubic
Automatically reconnects dropped chat response streams and resume from the last received chunk so turns keep rendering without duplication. Tightens stall detection and probes on tab visibility to prevent freezes like chat#1923.

- **Bug Fixes**
  - Added `useStreamRecovery` to detect silence and call `resumeStream()`; also checks on `visibilitychange` and polls every 3s with a 10s stall window and 8s cooldown.
  - Set `resume: true` in `useChat` to reattach to in‑progress streams on mount.
  - Implemented gap‑free resume by counting received SSE chunks via `lib/chat/createChunkCountingFetch` and sending `startIndex=last+1`; `prepareReconnectToStreamRequest` uses `lib/chat/buildStreamReconnectUrl` to rebuild `${api}/${chatId}/stream`, reads the server chat id from a ref, attaches bearer auth, and resets the position on fresh reads and chat switches to avoid stale indices. Unit tests cover both helpers and the decision logic.
  - Introduced `shouldRecoverStalledStream` (unit‑tested) as the pure decision logic.
  - Typed the recovery ref to remove an unused placeholder param flagged by lint.

- **Migration**
  - Requires the resume route `GET /api/chat/{chatId}/stream` from `api#809` (per `docs#286`) to be available.

<sup>Written for commit ea642b3edc90c30da29963db5a31f9fb6ef0ce84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1924?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Chat responses now automatically reconnect when streaming stalls.
  * In-progress responses resume from the last received content, reducing duplicate or lost text.
  * Recovery checks immediately when returning to an inactive browser tab.
* **Bug Fixes**
  * Improved handling of interrupted or silently disconnected chat streams.
  * Added safeguards to prevent overlapping recovery attempts and repeated reconnects.

### DIFF
--- a/hooks/useChatTransport.ts
+++ b/hooks/useChatTransport.ts
@@ -119,7 +119,7 @@ export function useChatTransport({
         // Reconnect hits `GET {api}/{chatId}/stream` — recoup-api's resume
         // route, which is authenticated like every other endpoint. Without
         // this the reconnect 401s and a dropped stream stays dropped.
-        prepareReconnectToStreamRequest: async ({ api, id }) => {
+        prepareReconnectToStreamRequest: async ({ api }) => {
           const accessToken = await getAccessToken().catch(() => null);
           const headers: Record<string, string> = {};
           if (accessToken) headers.Authorization = `Bearer ${accessToken}`;
@@ -129,8 +129,15 @@ export function useChatTransport({
           // `api` of our own. Returning one replaces the whole URL, so the
           // path has to be rebuilt, not appended to: appending the query to
           // the base produced `POST`-only `/api/chat?startIndex=N` and 405s.
+          //
+          // Built from `chatIdRef`, NOT the `id` the callback is handed: that
+          // is the `useChat` INSTANCE id, which for a new chat is still the
+          // client placeholder while the api-minted id lives in the ref. Using
+          // it reconnected to a chat that does not exist and 404'd. This is the
+          // same ref the request body already sends as `chatId`.
           const last = lastChunkIndexRef.current;
-          const url = `${api}/${id}/stream${last === null ? "" : `?startIndex=${last + 1}`}`;
+          const chatId = chatIdRef.current;
+          const url = `${api}/${chatId}/stream${last === null ? "" : `?startIndex=${last + 1}`}`;
 
           return { headers, api: url };
         },
--- a/hooks/useChatTransport.ts
+++ b/hooks/useChatTransport.ts
@@ -2,6 +2,8 @@ import { useMemo, useCallback, useRef } from "react";
 import { DefaultChatTransport } from "ai";
 import { getClientApiBaseUrl } from "@/lib/api/getClientApiBaseUrl";
 import { usePrivy } from "@privy-io/react-auth";
+import { createChunkCountingFetch } from "@/lib/chat/createChunkCountingFetch";
+import { buildStreamReconnectUrl } from "@/lib/chat/buildStreamReconnectUrl";
 
 interface UseChatTransportOptions {
   /** Chat row id (also the AI SDK `useChat({ id })` instance id). */
@@ -35,15 +37,13 @@ export function useChatTransport({
   // stable while always sending the ids current as of the request.
   const chatIdRef = useRef(chatId);
   const sessionIdRef = useRef(sessionId);
-  // Absolute index of the last stream chunk this client has actually received.
-  // Drives `startIndex` on reconnect so a resume is gap-free rather than a
-  // replay. Counted from the wire, NOT taken from
-  // `x-workflow-stream-tail-index`: that header reports the tail at the moment
-  // the read was opened, so a read that stays open past it under-reports (a
-  // live read returning 22 chunks advertised a tail of 9). The header is only
-  // a base for the read's starting position; what we have actually consumed is
-  // that base plus the chunks counted off the body.
+  // Absolute index of the last stream chunk this client received, maintained
+  // by `createChunkCountingFetch`. Drives `startIndex` on reconnect.
   const lastChunkIndexRef = useRef<number | null>(null);
+  // Switching chats voids the position: the transport is memoised for the
+  // lifetime of the hook, so without this a reconnect for the new chat could
+  // resume at an index belonging to the old one.
+  if (chatIdRef.current !== chatId) lastChunkIndexRef.current = null;
   chatIdRef.current = chatId;
   sessionIdRef.current = sessionId;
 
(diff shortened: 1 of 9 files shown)


sweetmantech:
## Refinements after comparing against upstream open-agents

Three changes in `3c909926` + `3cb9682f`.

### 1. Consume `x-workflow-stream-tail-index` — the gap I flagged as unfixable

I originally wrote that precise `startIndex` was impossible because `useChat` exposes no chunk count. **That was wrong** — the server can just tell us, which is what upstream does via `readable.getTailIndex()`. [api#809](https://github.com/recoupable/api/pull/809) now returns that header, so:

- a custom transport `fetch` captures `x-workflow-stream-tail-index` off every response
- `prepareReconnectToStreamRequest` sends `startIndex = tail + 1` (the header is 0-based, so the next unseen chunk is tail + 1)

Reconnects are now **gap-free** rather than replaying the turn from chunk zero.

### 2. Thresholds tightened

| | before | now | upstream |
|---|---|---|---|
| stall window | 20 s | **10 s** | n/a (see below) |
| cooldown | 15 s | **8 s** | 8 s (`STREAM_RECOVERY_MIN_INTERVAL_MS`) |
| poll | 5 s | **3 s** | n/a |

Safe *because* of (1): an unnecessary reconnect now costs a request rather than re-rendering content the client already has.

**I did not adopt upstream's `STREAM_RECOVERY_STALL_MS = 4_000`, and it is worth saying why.** That constant feeds a scheduler their `shouldScheduleStallRecovery` unconditionally disables:

```ts
export function shouldScheduleStallRecovery(options: {…}): boolean {
  void options;
  return false;
}
```

So it is **not a live stall threshold** to copy — upstream has no stall-based recovery at all. It also would not survive our workload: a single legitimate tool call streams nothing for up to **~200 s** (measured on the prod repro), so a 4 s window would fire dozens of pointless reconnects per turn.

### 3. Visibility probe

Added, and it is upstream's *only* live recovery trigger. A backgrounded tab can have its connection killed silently, and no amount of waiting produces a chunk to time out on — so a visibility check skips the silence window. The cooldown still applies, so a focus-flapping tab cannot spam reconnects.

## Why I kept the stall trigger rather than adopting upstream's posture wholesale

Upstream's live triggers are `status === "error"` and a visibility probe when `status === "ready"`. **Our failure mode produces neither**: the stream ends with a clean `[DONE]`, so there is no error, and the prod repro had the user watching a **visible** tab for nine minutes. Adopting upstream's posture exactly would have left our own bug unfixed except on tab re-focus.

## Status

- **366 chat tests passing** (3 new cases covering the visibility path and its cooldown).
- `tsc --noEmit`: 7 errors, identical to the `main` baseline, zero in files this PR touches.
- `eslint`: the one error that was mine (unused placeholder param in a `useRef` initializer) is fixed in `3cb9682f`. The remaining 5 in `hooks/` are pre-existing on `main` — including `useVercelChat.ts` `'authenticated' is assigned a value but never used`, which is line 83 on `main` and only shifted to 84 by this diff.

Preview verification against a real dropped stream is still owed.

sweetmantech:
## Preview verification — the reconnect works end to end

Signed in as the real account on preview `chat-ord57m1sv-recoup.vercel.app` (`bba6a96b`), pointed at `test-recoup-api` which now carries the merged [api#809](https://github.com/recoupable/api/pull/809) route. Prompt: five sequential 45-second bash calls, with `fetch` instrumented to log every `/api/chat` request and count SSE frames off a tee of the body.

| request | window | status | chunks |
|---|---|---|---|
| `POST /api/chat` | 17:09:45 → 17:11:48 (**123 s**) | 200 | 418 |
| `GET /api/chat/2daee8a8…/stream?startIndex=348` | 17:10:28 → 17:12:28 | **200** | **72** |

The main stream died at **123 s** again — third independent reproduction, on a third account and deployment. The reconnect resumed at index 348 and carried the turn to completion.

**The turn finished with no duplication**, which is the assertion that matters for a resume:

```
alpha: 1   bravo: 1   charlie: 1   delta: 1
```

Each word rendered exactly once. (`echo` appears 6× because the word is in all five bash commands plus the answer.) A wrong `startIndex` would have shown these two or more times.

reconnected turn completing

## It took three runs, and each found a different bug

| commit | reconnect URL | result |
|---|---|---|
| `cb8cea2e` | `/api/chat?startIndex=376` | **405** — appended the query to the base, destroying the path |
| `aa74c4f2` | `/api/chat/ca264e2f…/stream?startIndex=141` | **404** — right shape, wrong chat id |
| `bba6a96b` | `/api/chat/2daee8a8…/stream?startIndex=348` | **200** ✅ |

1. **405** — `prepareReconnectToStreamRequest` receives `api` as the *base*; the SDK only falls back to `${api}/${id}/stream` if the callback returns no `api`. Returning one replaces the whole URL, so appending a query to the base produced a GET against the POST-only chat endpoint.
2. **404** — the `id` the callback receives is the `useChat` **instance** id, still the client placeholder for a new chat. The api-minted id lives in `chatIdRef` (which exists precisely because `useChat` captures the transport at mount and never swaps it). Now uses that ref — the same value the request body already sends as `chatId`.

Worth stating: **all 86 test files passed against every one of those broken URLs.** The tests cover the *decision* to reconnect, not the URL the SDK ultimately builds. Nothing but a live preview run would have caught these.

Consistent across all three runs: the stall detector fired reliably, and the chunk counter tracked real positions (376 → 409, 141 → 180 → 209, 348).

## Two findings worth acting on separately

**1. `x-workflow-stream-tail-index` is invisible to browser clients.** My log read it as `null` on the 200 reconnect. `getCorsHeaders()` sets `Access-Control-Allow-Headers` but no **`Access-Control-Expose-Headers`**, so cross-origin JS cannot read custom response headers — `x-workflow-run-id` is equally hidden. This PR does not depend on it (it counts frames off the wire instead), but the header added in api#809 is currently unusable from the browser, and would silently break the SDK's `WorkflowChatTransport` if we ever adopt it. Needs a small api follow-up.

**2. The reconnect fired at 17:10:28, before the original stream died at 17:11:48.** With a 10 s stall window and 45 s sleeps, legitimate tool-call silence trips the detector, so we briefly ran two concurrent reads of the same stream. Harmless here — resuming from 348 produced no duplication — but it is wasted work. If we want to avoid it, the lever is upstream's approach of probing whether the server still considers the stream live before reconnecting, rather than reconnecting blind. Flagging rather than changing it now, since the current behaviour is correct.

hooks/useChatTransport.ts
```diff
@@ -61,6 +70,77 @@ export function useChatTransport({
...
+          // client placeholder while the api-minted id lives in the ref. Using
+          // it reconnected to a chat that does not exist and 404'd. This is the
+          // same ref the request body already sends as `chatId`.
+          const last = lastChunkIndexRef.current;
+          const chatId = chatIdRef.current;
+          const url = `${api}/${chatId}/stream${last === null ? "" : `?startIndex=${last + 1}`}`;
+
+          return { headers, api: url };
+        },
       }),
```
sweetmantech: OCP
- actual: net new code added inline to useChatTransport
- required: new lib file for the net new code imported into useChatTransport

sweetmantech:
## Both review comments addressed in `ea642b3e`

### OCP — net new logic was inline in the hook

Extracted, so `useChatTransport` is wiring again and both units are directly testable:

| file | LOC | role |
|---|---|---|
| `lib/chat/createChunkCountingFetch.ts` | 89 | wraps `fetch`, counts SSE frames off a tee of the body, reports the absolute position |
| `lib/chat/buildStreamReconnectUrl.ts` | 25 | pure URL builder |
| `hooks/useChatTransport.ts` | 101 | wiring only |

### Stale index — valid, and I've closed it at both ends

The concern is real. It is also narrower than it first reads, which is worth stating precisely rather than just agreeing: the counter **re-seeds from each request's own `startIndex`**, so it self-corrects on the first frame of any new read. A stale value can only be sent in the window between issuing a request and its first frame arriving.

That window is not theoretical. With a 10 s stall threshold and a slow sandbox start, a reconnect can land in it — and then `startIndex` points past chunks the client never saw, which is a silent gap rather than a visible error. Worse than a replay.

Closed at both ends:

1. **`createChunkCountingFetch` reports `null` the moment it issues a read with no `startIndex`** — i.e. one starting from chunk zero — *before* awaiting the response. Covers a new turn and a new chat's first POST. The test asserts the reset lands synchronously, since a reset that only arrived with the first frame would leave the window open.
2. **`useChatTransport` clears the ref when `chatId` changes**, because the transport is memoised for the lifetime of the hook and would otherwise carry a position across chats.

Note the comment referenced `tailIndexRef`; that ref was renamed to `lastChunkIndexRef` in `cb8cea2e` when the implementation moved from trusting the response header to counting the wire. Same object, same concern.

### Coverage

9 new unit tests: frame counting, `[DONE]` exclusion, seeding from `startIndex`, **both** reset paths, non-reset on a genuine resume, body pass-through, and bodyless (204) responses.

**88 chat test files passing**; `tsc --noEmit` delta 0 vs `main`; `eslint` clean.

Re-running the live preview verification on this head before I'd call it done — the last three rounds each found something unit tests could not.

PR #1925, opened 2026-08-03
fix(chat): keep probing after the turn stops looking in-flight

sweetmantech:
Row 1 of [chat#1923](https://github.com/recoupable/chat/issues/1923) — **the original bug is live on prod**. [chat#1924](https://github.com/recoupable/chat/pull/1924) shipped the plumbing but the trigger never fires in the real failure mode.

## Reproduced on prod after #1924 merged

Original prompt, prod `7a56bb65`:

| | |
|---|---|
| `POST /api/chat` | 19:13:42 → **19:15:45** (123 s), 1,498 chunks |
| Reconnect requests | **zero** |
| Run kept writing until | **19:19:00** — 3¼ min later |
| Persisted server-side | 25 parts (6 `step-start`, 6 `reasoning`, 3 `text`, 2 `tool-skill`, 8 `tool-bash`) |
| UI | frozen, no stop control, composer idle |

## Root cause

`shouldRecoverStalledStream` gated on `IN_FLIGHT_STATUSES = {streaming, submitted}`. The failure ends the stream **cleanly** — `[DONE]` with no `finish` chunk — so `useChat` transitions **out** of in-flight and the interval stops evaluating. Everything downstream (frame counting, URL building, auth, the api route) was correct and simply never got invoked.

**The earlier verification passed for the wrong reason.** Its prompt was five sequential 45 s sleeps, which guarantees a stall *while still streaming* — a resume was already in flight at 17:10:28 before the stream died at 17:11:48. It never exercised the post-drop path. The original prompt is many quick tool calls, so nothing was in flight when the stream ended.

This is the case upstream open-agents covers with `status === "ready"` → probe. I reviewed that, called it redundant because our resume route self-probes via its 204, and dropped it. Wrong: **the route cannot probe itself if nothing calls it.**

## Fix

A cut-off stream and a completed one are indistinguishable from the client — both end with `[DONE]`. So stop trusting the appearance and ask the server.

- **`POST_TURN_PROBE_MS`** (5 min): after the in-flight → not-in-flight edge, keep evaluating under the existing cooldown.
- **204 ends it.** `createChunkCountingFetch` reports a 204 on the resume path — the server saying there is nothing to resume, the only authoritative end-of-turn signal. A normally-completed turn therefore costs **one extra request**, then stops.
- **Any chunk re-arms**, so a resumed stream that drops again is caught.

The window is a backstop, not the control: the 204 is what normally ends probing. It bounds the pathological case where that answer never arrives.

This also subsumes the "reconnect fired before the stream died" waste flagged on #1924, since the trigger is no longer purely silence-based.

## Tests

5 new cases on the policy, RED before GREEN — the first one reproduces the prod bug directly:

- recovers after the turn leaves in-flight, while the window is open
- stops once the window expires
- never probes a turn that never ran (so idle chats don't poll)
- stops once the server has confirmed nothing to resume
- honours the cooldown on post-turn probes

Plus 2 on the fetch: reports no-active-stream on a 204 resume, and does **not** for a normal streaming response.

**88 chat test files passing**; `tsc --noEmit` clean in every touched file; `eslint` clean apart from the pre-existing `useVercelChat` `'authenticated' is assigned but never used`, which is on `main`.

## Verification owed

**This must be verified against the original prompt on a preview before merge — not a sleep-based scenario.** That substitution is exactly what hid the bug last time. Results will be posted here.

---
## Summary by cubic
Keep probing for a stalled stream after a turn exits in-flight, and stop immediately when the server returns 204. Fixes the prod bug where streams end with `[DONE]` (no finish chunk) and the UI freezes while the run continues.

- **Bug Fixes**
  - Added a 5‑min `POST_TURN_PROBE_MS` window after leaving in-flight; respects existing cooldown.
  - Treat 204 from the resume route as the authoritative end-of-turn; `createChunkCountingFetch` signals no active stream and `useStreamRecovery` halts.
  - Track the in-flight → not-in-flight edge; any new chunk re-arms recovery. Idle chats never poll; probing stops once the server confirms done.
  - Tests cover post-turn probing, cooldowns, and the 204 path.

<sup>Written for commit fef0bd405ded3f1c88e7bc0ba85b0c01c2138748. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1925?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/hooks/useChatTransport.ts
+++ b/hooks/useChatTransport.ts
@@ -26,6 +26,9 @@ export function useChatTransport({
   chatId,
   sessionId,
 }: UseChatTransportOptions) {
+  // Flipped when the resume route answers 204. Read by the recovery hook to
+  // stop probing a turn the server says is over.
+  const noActiveStreamRef = useRef(false);
   const { getAccessToken } = usePrivy();
   const baseUrl = getClientApiBaseUrl();
 
@@ -97,5 +100,5 @@ export function useChatTransport({
     [baseUrl, getAccessToken],
   );
 
-  return { transport, getHeaders };
+  return { transport, getHeaders, noActiveStreamRef };
 }
--- a/hooks/useStreamRecovery.ts
+++ b/hooks/useStreamRecovery.ts
@@ -11,6 +11,12 @@ interface UseStreamRecoveryOptions {
   /** Current `useChat` status. */
   status: string;
   /**
+   * True once the resume route has answered 204 for this turn. Ends probing —
+   * the server is the only party that can distinguish a finished turn from a
+   * cut-off one.
+   */
+  serverConfirmedDone: boolean;
+  /**
    * Changes whenever the assistant produces output. Any value that moves as
    * chunks arrive works — the hook only reads it to timestamp "last activity".
    */
@@ -35,12 +41,18 @@ interface UseStreamRecoveryOptions {
  */
 export function useStreamRecovery({
   status,
+  serverConfirmedDone,
   activityMarker,
   resumeStream,
 }: UseStreamRecoveryOptions): void {
   const lastChunkAtRef = useRef<number | null>(null);
   const lastRecoveryAtRef = useRef(0);
   const isRecoveryInFlightRef = useRef(false);
+  // When the turn last stopped looking in-flight. Opens the post-turn probe
+  // window — the stream ends identically whether the turn finished or was cut
+  // off, so the client cannot tell them apart without asking.
+  const turnEndedAtRef = useRef<number | null>(null);
+  const wasInFlightRef = useRef(false);
 
   // Timestamp activity. Kept in a ref so the polling effect never re-registers
   // as output streams in.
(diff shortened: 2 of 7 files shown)


sweetmantech:
Superseded by [chat#1927](https://github.com/recoupable/chat/pull/1927) — **do not merge this one.**

Preview testing of this branch (`chat-1eerhhswj`, original prompt) showed the fix works but the *shape* does not: **24 reconnects, 12 of them throwing, 12,146 chunks re-downloaded for a ~4,000-chunk turn**, still retrying five minutes after the run ended.

The defect is not in this PR's new trigger specifically — it is that a polling cadence was attached to `resumeStream()`, which opens a stream rather than answering a question. Adding a third trigger to two existing ones made that unavoidable.

#1927 replaces all three with one rule gated on `ChatSummary.isStreaming` from `GET /api/sessions/{sessionId}/chats` — an endpoint that already existed and that I wrongly reported as missing. It is a **net deletion** (245+/246-) and removes the storm by construction rather than by tuning a cooldown.

Closing recommended so nobody merges this by mistake.

sweetmantech:
Closing — superseded and never safe to merge.

Preview testing showed this shape storms: **24 reconnects in six minutes, 12 throwing, 12,146 chunks re-downloaded** for a turn of ~4,000. Root cause was attaching a polling cadence to `resumeStream()`, which *opens a stream* rather than answering a question.

Everything here is now handled by [chat#1935](https://github.com/recoupable/chat/pull/1935), which replaced the hand-rolled recovery with `WorkflowChatTransport` — a single generator looping `while (!gotFinish)`, with no trigger to miss and no ref to hold. Verified on prod: the original prompt streams 456 s and settles cleanly.

The reasoning is preserved in the dated callouts on [chat#1923](https://github.com/recoupable/chat/issues/1923).

PR #1926, opened 2026-08-03
feat(chat): resume a refreshed page from the recent tail, not chunk zero

sweetmantech:
Row 6 of [chat#1923](https://github.com/recoupable/chat/issues/1923). **Blocked by [api#810](https://github.com/recoupable/api/pull/810)** — see below.

## The problem

A reload mid-turn has no counted position, so `resume: true` reconnects with no `startIndex` and the server replays the **entire** response — 418 chunks in the verified prod run. All of that content is already persisted and rendered from the database on load, so the replay is pure waste.

## The fix

Ask for a bounded tail instead: `REFRESH_RESUME_WINDOW = -50`. The server resolves negative indices relative to the end of the stream and reports where the read landed via `x-workflow-stream-tail-index`, which is what lets later retries in the same session return to exact absolute positions.

**A counted position always wins over the window.** The fresh-load flag is cleared the moment any chunk arrives, so only the *first* reconnect of a freshly loaded page uses it — every subsequent one resumes from what the client actually received.

## Hard dependency on api#810

The tail header is **invisible to browser JS** until `Access-Control-Expose-Headers` ships. Per the SDK docs, without that header a negative `startIndex` is *"assumed to be 0, replaying the entire stream"* — so merging this first would be a no-op at best. **api#810 must land first.**

## Why this doesn't remove the frame counting

Worth restating, because it is the thing most likely to be misread: headers are sent **before** the body, so `x-workflow-stream-tail-index` reports the tail at the moment the read was *opened*, never where it ended. A read advertising `tail: 9` went on to deliver 22 chunks. Counting frames off the wire remains the only way to know how far a long-lived read actually got; this header is an anchor for *relative* indices and nothing more.

## Tests

4 new cases, RED before GREEN: a fresh load requests the bounded window; the window is negative; a counted position is preferred over it; a non-fresh reconnect with no position still reads from the beginning.

**88 chat test files passing**; `tsc --noEmit` clean in touched files; `eslint` clean.

## Merge order

[api#810](https://github.com/recoupable/api/pull/810) → this. Note it also touches `hooks/useChatTransport.ts`, as does [chat#1925](https://github.com/recoupable/chat/pull/1925); whichever lands second will want a rebase.

---
## Summary by cubic
Resume a refreshed chat page from the recent tail instead of chunk 0 to avoid replaying entire responses and speed up reloads. Applies only to the first reconnect when there’s no counted position; counted positions still win.

- **New Features**
  - On fresh reloads with no position, request `startIndex=-50` to fetch only the recent tail.
  - If a counted position exists, resume from `lastChunkIndex + 1`.
  - Use `x-workflow-stream-tail-index` to anchor subsequent retries; frame counting stays for progress.
  - Clear the fresh-load flag once any chunk arrives so later reconnects use the counted position.

- **Dependencies**
  - Blocked by `api#810` to expose `x-workflow-stream-tail-index` via `Access-Control-Expose-Headers`.
  - Merge order: `api#810` → this.

<sup>Written for commit 8000640805413cd056314b51903003bfa218f2ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/hooks/useChatTransport.ts
+++ b/hooks/useChatTransport.ts
@@ -40,6 +40,10 @@ export function useChatTransport({
   // Absolute index of the last stream chunk this client received, maintained
   // by `createChunkCountingFetch`. Drives `startIndex` on reconnect.
   const lastChunkIndexRef = useRef<number | null>(null);
+  // True until this transport has seen any stream of its own. A reconnect
+  // while it is set is a fresh page load resuming a turn it never watched, so
+  // it asks for a bounded tail rather than replaying the whole response.
+  const isFreshLoadRef = useRef(true);
   // Switching chats voids the position: the transport is memoised for the
   // lifetime of the hook, so without this a reconnect for the new chat could
   // resume at an index belonging to the old one.
@@ -74,6 +78,10 @@ export function useChatTransport({
           baseUrl,
           onPosition: (index) => {
             lastChunkIndexRef.current = index;
+            // Once a chunk lands, this transport has watched a turn of its
+            // own — later reconnects resume from the counted position rather
+            // than asking for a blind tail.
+            if (index !== null) isFreshLoadRef.current = false;
           },
         }),
         // Reconnect hits recoup-api's resume route, which is authenticated
@@ -86,11 +94,9 @@ export function useChatTransport({
 
           return {
             headers,
-            api: buildStreamReconnectUrl(
-              api,
-              chatIdRef.current,
-              lastChunkIndexRef.current,
-            ),
+            api: buildStreamReconnectUrl(api, chatIdRef.current, lastChunkIndexRef.current, {
+              isFreshLoad: isFreshLoadRef.current,
+            }),
           };
         },
       }),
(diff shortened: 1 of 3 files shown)


sweetmantech:
Closing — the **goal is still valid, the code is dead**.

This PR modifies `buildStreamReconnectUrl.ts` and its test, both of which no longer exist: [chat#1935](https://github.com/recoupable/chat/pull/1935) deleted the hand-rolled reconnect layer when it adopted `WorkflowChatTransport`. Rebasing would mean rewriting it from scratch.

**The problem it targets is real and still unfixed.** I verified on `main`: `createWorkflowChatTransport` does not set `initialStartIndex`, so refreshing a page mid-turn resumes at chunk 0 and replays the entire turn.

The fix is now far smaller than this PR. `WorkflowChatTransport` supports a negative `initialStartIndex` natively — it reads `x-workflow-stream-tail-index` and resolves the absolute position itself — so this becomes roughly one option:

```ts
initialStartIndex: -50,   // on refresh, read the recent tail, not all ~4,000 chunks
```

Tracked as a follow-up rather than carried here.

PR #1927, opened 2026-08-03
refactor(chat): port stream recovery to upstream open-agents' event-driven shape

sweetmantech:
Replaces the three-trigger recovery with **one rule**, and deletes more than it adds. Supersedes [chat#1925](https://github.com/recoupable/chat/pull/1925) — please close that rather than merge it.

## Why the previous shape failed

Preview `chat-1eerhhswj` running the original prompt:

| | |
|---|---|
| `POST /api/chat` | died at **124 s**, 1,236 chunks |
| First reconnect | 20:04:59 → **200**, correct |
| Then | **24 reconnects** — 12 succeeded, 12 threw |
| Re-downloaded | **12,146 chunks** for a ~4,000-chunk turn |
| Still retrying at | five minutes **after** the run ended |

The defect was not in any single trigger. It was attaching a polling cadence to **`resumeStream()`, which opens a stream rather than answering a question**. Three triggers stacked on that produced behaviour none of them had alone.

## The fix

A dropped stream and a finished one are **indistinguishable from the client** — both end with `[DONE]` and no `finish` chunk, and `useChat` leaves `streaming` either way. So stop inferring, and ask:

```
every 5s: if (not receiving && server says isStreaming) resumeStream()
```

## The probe already existed — my earlier claim was wrong

I previously said `active_stream_id` was "surfaced by no endpoint" and planned a docs + api pair to add one. **That was wrong.** I grepped `lib/chats/`, `lib/supabase/chats/` and `app/api/chats/`, and missed `lib/sessions/chats/`.

`GET /api/sessions/{sessionId}/chats` has always returned `ChatSummary.isStreaming`, derived from `chats.active_stream_id`, **carrying no message bodies** — [`getChatSummaries.ts` L67](https://github.com/recoupable/api/blob/main/lib/sessions/chats/getChatSummaries.ts#L67), documented as `ChatSummary` in `sessions.json`. It is also precisely the endpoint upstream open-agents probes. So a three-PR plan collapsed to this one, with **no server change at all**.

The sibling `GET /api/sessions/{sessionId}/chats/{chatId}` also reports `isStreaming` but bundles the full transcript — the wrong one to poll, and worth saying out loud so nobody switches to it later.

## Deleted

- `lib/chat/shouldRecoverStalledStream.ts` **and its test file**
- the stall-on-silence branch and `STREAM_STALL_MS`
- the visibility branch
- the `activityMarker` plumbing through `useVercelChat`

## Added

- `lib/chat/shouldResumeStream.ts` — the single rule
- `lib/chat/fetchChatIsStreaming.ts` — the probe, returning **`null`** when the probe itself failed, so an unanswered question is never read as a "no"
- a **failure ceiling** (3 consecutive), so a run that has gone away cannot retry forever — the second half of what went wrong last time

**Net `245 insertions, 246 deletions`.**

## Tests

6 cases, RED before GREEN — each maps to an observed failure: resumes on a `yes`; **does not** on a `no`; **does not while chunks are arriving** (the concurrent-read storm); no stacked attempts; honours the cooldown; **gives up after repeated failures** rather than retrying forever.

**88 chat test files passing**; `tsc --noEmit` clean in every touched file; `eslint` clean apart from the pre-existing `useVercelChat` `'authenticated' is assigned but never used`, which is on `main`.

## Verification owed

Must be verified on a preview **with the original prompt** — many quick tool calls, not long sleeps. That substitution is what produced a false pass before. The specific new assertion: **zero reconnects once the probe reports `isStreaming: false`**.

---
## Summary by cubic
Ports chat stream recovery to open‑agents’ event‑driven design, adds a transport‑level stream‑end trigger, and gates actions on the server’s `isStreaming`. On stream end, it now either resumes once or ends the turn if the run is gone; recovery can fire again while a prior resume is still streaming.

- **Bug Fixes**
  - Recovery runs on `visibilitychange`, `focus`, `online`, and end‑of‑body with an 8s minimum between actions; no polling. Stream‑end is fired from `createChunkCountingFetch` and routed through `useChatTransport` to `useStreamRecovery`.
  - Probes `GET /api/sessions/{sessionId}/chats` on `[0ms, 1s, 2.5s, 5.5s, 10s]`; resumes on the first “yes”; retries immediately on `status === "error"`. If all probes say “not streaming” at stream end, calls `stop()` to settle the turn.
  - Releases the probe-in‑flight flag before calling `resumeStream`, so a later stream end can still trigger recovery while the resumed stream is open.

- **Refactors**
  - `useStreamRecovery` reads auth from `usePrivy` in `@privy-io/react-auth`; callers no longer pass tokens.
  - Extracted `wait()` to `lib/async/wait.ts`; `onStreamEndRef` is optional for one‑shot transports.

<sup>Written for commit 7a01f395a1b6b41c8ff176bb3f96470d6f862713. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1927?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat stream recovery when returning to the app, regaining focus, reconnecting online, or completing a stream.
  * Recovery now verifies that the server still has an active stream before resuming.
  * Added retry handling for resume errors and temporary connectivity or server-check failures.
  * Removed polling and stall-based recovery to reduce unnecessary recovery attempts.
  * Stream completion now triggers timely recovery checks when appropriate.


sweetmantech:
## Preview verification — `75fc5b32` on `chat-git-fix-probe-gated-stream-recovery-recoup`

Signed in as sweetman@recoupable.com, original prompt (*"can you send me an email with the status of all artists across accounts?"*), `fetch` instrumented to count probes and reconnects separately. Build confirmed new by fingerprint (`b6ab7d9b…`, previous `57bb2dba…`).

### Results

| # | Assertion | Observed | |
|---|---|---|---|
| 1 | No polling | **263 s focused, zero probes / zero reconnects**; second silent window of **100 s**, also zero | ✅ |
| 2 | One event → one probe → one reconnect | focus at 303.6 s → probe `200` → reconnect `200` at 304.4 s | ✅ |
| 3 | Overlapping events collapse | 4 events (`focus`,`online`,`focus`,`visibilitychange`) inside the 8 s window → **probes stayed at 1** | ✅ |
| 4 | Nothing fires after the run ends | post-run focus → resume returned **`204`** (the no-op contract), no stream opened | ✅ |

Against the polled version on the previous build: **16 reconnects across two drops, firing at the 5 s probe interval**. Here the same turn produced **one probe and one content-bearing reconnect**.

### Notes, including what this does *not* fix

- **The focused-tab drop is still not recovered, by design.** No `visibilitychange`/`focus`/`online` fires while you sit on the tab, so nothing triggers. Upstream open-agents has the same gap. That case is the ~123 s ceiling and is fixed server-side, not here — it is now the highest-priority row on chat#1923.
- **The `retry-error` path resumes without probing** (`status === "error"` → direct `resumeStream()`), faithful to upstream's policy. Observed twice; both were cheap — the second returned `204`. Bounded by the same 8 s interval.
- **The visibility transition was synthesized.** CDP tab-selection does not toggle `document.visibilityState`, so the events were dispatched directly at the listeners the browser would drive. What is verified end-to-end is the handler → probe → resume chain against the live server, not the browser's event emission.
- **This run did not send an email**, so it does not reproduce Nena's case end to end. The agent's sandbox failed auth against the api (`401 "Failed to verify authentication token"`) — unrelated to this PR, which touches only client reconnect logic.

### Checks

376 chat tests pass (11 new RED→GREEN, including a 120 s *"never fires without an event"* guard), `tsc --noEmit` clean in touched files, eslint clean.

sweetmantech:
## Preview verification — `ee4d44dd`, original prompt, untouched tab

**The turn completed and the email was sent.** First end-to-end success; every prior attempt froze at the two-minute mark.

Signed in as sweetman@recoupable.com on `chat-git-fix-probe-gated-stream-recovery-recoup`. No synthetic events — the tab was left alone for the whole run.

### Reconnect chain

| # | Probe | Answer | Reconnect | `startIndex` |
|---|---|---|---|---|
| 1 | 160.0 s | `true` | `200` @ 161.0 s | 1592 |
| 2 | 281.8 s | `true` | `200` @ 282.6 s | 2986 |
| 3 | 403.2 s | `true` | `200` @ 403.8 s | 4257 |
| 4 | 524.5 s | `true` | `200` @ 525.6 s | 5766 |
| 5 | 646.2 s | `true` | `200` @ 647.2 s | 6963 |

Five reconnects, each **~121 s** apart — the connection cap measured in chat#1928 — each resuming from the exact chunk the client last received, with no replay and no gap. The turn ran **~13 minutes**, about 6× the old ceiling, and rendered to completion.

### Final output

The assistant delivered the full report and confirmed **"Email sent"**: 98 connected social profiles across the roster, a platform breakdown (Spotify 29, Instagram 27, TikTok 13, Twitter/X 11, YouTube 6, Facebook 4, Apple Music 3, Threads 1), and flagged data caveats rather than papering over them (4 org artists with no socials, mismatched label records, a null follower count).

### Against the earlier attempts

| Build | Result |
|---|---|
| polled recovery | 16 reconnects for one turn, 5 s cadence |
| event-driven only | 0 reconnects on a focused tab — froze at ~173 s |
| status-edge trigger | 1 reconnect, then froze |
| single-probe stream-end | 2 reconnects; a false `isStreaming: false` at 410 s ended recovery |
| **this build** | **5 reconnects, chained, turn completed, email sent** |

### Checks

386 tests pass (5 new RED→GREEN for the backoff), `tsc --noEmit` clean unfiltered, eslint clean on touched files, `next build` compiles.

hooks/useStreamRecovery.ts
```diff
@@ -1,109 +1,132 @@
...
  */
 export function useStreamRecovery({
+  sessionId,
+  chatId,
   status,
-  activityMarker,
   resumeStream,
-}: UseStreamRecoveryOptions): void {
-  const lastChunkAtRef = useRef<number | null>(null);
+  getAccessToken,
```
sweetmantech: KISS
- why is getAccessToken passed as a prop rather than directly imported from the provider?

hooks/useVercelChat.ts
```diff
@@ -233,12 +238,15 @@ export function useVercelChat({
...
+  // A long turn's SSE stream ends before the run does — connections are capped
+  // at ~120s (chat#1928). Ask the server at each stream end and resume from the
+  // last chunk received if the run is still live.
+  onStreamEndRef.current = useStreamRecovery({
+    sessionId,
+    chatId: transportChatId,
     status,
-    activityMarker: messages,
     resumeStream,
+    getAccessToken,
```
sweetmantech: KISS - Why do we need to re-export the getAccessToken function? It already hs a provicer, right?

lib/chat/probeChatIsStreaming.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface ProbeChatIsStreamingOptions {
+  sessionId: string;
+  chatId: string;
+  getAccessToken: () => Promise<string | null>;
+  /** Override the schedule; `[0]` asks exactly once. */
+  delaysMs?: number[];
+}
+
+const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
```
sweetmantech: SRP - actual: wait defined in lib/chat/probeChatIsStreaming.ts should be defined in standalone lib.

hooks/useStreamRecovery.ts
sweetmantech: Fixed in `7f504929` — the hook now pulls `getAccessToken` from `usePrivy()` itself, matching `useChatTransport`. Only instance-scoped values (`sessionId`, `chatId`, `status`, `resumeStream`) are passed in now.

hooks/useVercelChat.ts
sweetmantech: Fixed in `7f504929` — the forwarding is gone. `useVercelChat` still uses `getAccessToken` for its own calls (file upload, model persistence), but no longer hands it to recovery.

lib/chat/probeChatIsStreaming.ts
sweetmantech: Fixed in `7f504929` — extracted to `lib/async/wait.ts`. Checked for an existing helper first; there wasn't one, so this is the first shared `wait`.

hooks/useStreamRecovery.ts
```diff
@@ -1,109 +1,116 @@
...
+      const { sessionId: sid, chatId: cid, status: st, resumeStream: resume } = stateRef.current;
+      if (!sid) return;
 
-    lastRecoveryAtRef.current = now;
-    isRecoveryInFlightRef.current = true;
+      // Cheap local rejections first — never spend a probe when the answer
+      // cannot change the outcome.
+      if (
+        isAttemptInFlightRef.current ||
+        st === "streaming" ||
```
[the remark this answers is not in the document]
sweetmantech: Not applicable to the path this describes — `isStreamEnd` returns before any status check:

```ts
if (isStreamEnd) return isProbeInFlight ? "none" : "probe";   // <- first
if (now - lastRecoveryAt < MIN_INTERVAL) return "none";
if (status === "error") return "retry-error";
if (isVisibilityRecovery && status === "ready") { ... }
```

So the dropped-stream case you describe — SSE closes with `[DONE]`, no finish chunk, `useChat` still reporting `streaming` — does reach the probe. It is triggered from the transport's end-of-body rather than from status precisely because status is not trustworthy there.

Verified live on preview `ee4d44dd`: five consecutive drops, each probed and reconnected (`startIndex` 1592 → 2986 → 4257 → 5766 → 6963), turn ran ~13 min and completed.

The `status === "ready"` requirement applies only to the visibility/focus path, where it is deliberate: reconnecting while chunks are genuinely arriving replays them on top of the live stream.

lib/async/wait.ts
sweetmantech, line 1: DRY - do we already have a shared `wait` function somewhere else?  IF this is the only one, ignore this comment.

sweetmantech: Checked properly — it is the only one, so ignoring per your note.

Three searches across the repo (excluding `node_modules`/`.next`):

| Search | Hits |
|---|---|
| `new Promise(resolve => setTimeout(…))` outside tests | 1 — `lib/async/wait.ts` |
| exported `sleep`/`delay`/`pause`/`wait`/`timeout` | 1 — `lib/async/wait.ts` |
| imports of a wait-like helper | 1 — `probeChatIsStreaming.ts` |

Three test files inline their own `setTimeout` flush helper, but those are local test scaffolding rather than a shared utility, so I left them alone rather than widening the PR.

sweetmantech:
Closing — superseded by [chat#1935](https://github.com/recoupable/chat/pull/1935), which **deleted every file this PR touches**.

This was the port of upstream open-agents' event-driven recovery (`useStreamRecovery` + `getStreamRecoveryDecision` + the probe backoff). It got close — 2 of 3 preview runs passed — but each fix exposed another unverified state in a machine built from React refs: a trigger that fired once, an in-flight flag held across a whole 120 s connection, a final stream end that fired nothing.

`WorkflowChatTransport` removes the class rather than the instances, and deleted `createChunkCountingFetch`, `buildStreamReconnectUrl`, `shouldRecoverStalledStream`, `useStreamRecovery` and their tests along with it.

The investigation is the lasting value and it lives on [chat#1923](https://github.com/recoupable/chat/issues/1923) and [chat#1928](https://github.com/recoupable/chat/issues/1928) — including the finding that `ChatSummary.isStreaming` reads `false` while a turn is actively streaming, which is why gating recovery on it was never going to be sound.

Issue #1929, opened 2026-08-04
Upgrade chat, marketing, and api to Next.js 16.3 (and restore the lint that next@16 broke)


sweetmantech:
> **Status 2026-08-05:** **All 8 upgrade PRs are merged and on production — `chat`, `marketing`, and `api` all run `next@16.3.0`.** [chat#1930](https://github.com/recoupable/chat/pull/1930) → [#1932](https://github.com/recoupable/chat/pull/1932) → [#1933](https://github.com/recoupable/chat/pull/1933) all landed, and [api#811](https://github.com/recoupable/api/pull/811) is on production. `chat/main` now reports `next@16.3.0`, `@types/react@^19.2.18`, `lint: eslint .`, and no `images.domains` key. The upgrade itself is **complete**. The catalog-songs auth bug this verification surfaced has moved to **[#1939](https://github.com/recoupable/chat/issues/1939)** — scoping it turned up two *unauthenticated write endpoints*, which outrank it. The three follow-ups below are the only things left here; they want their own issues so this one can close. Two things surfaced that this issue did not anticipate when it was written: the [type-checker blocker](#type-checker-blocker) below, and the fact that **`pnpm lint` was a dead command in both `chat` and `marketing`** — not degraded, dead, linting zero files since the Next 16 upgrade.

Tracking issue for upgrading the three Next.js frontends — `chat`, `marketing`, and `api` — from `next@16.0.10` to `next@16.3.0` ([release notes](https://nextjs.org/blog/next-16-3), published 2026-08-03). All three are already on the v16 major, so this is a minor bump with no v15→v16 breaking changes left to absorb. The work that *is* here is the debt v16 left behind in these repos: `next lint` was removed in v16 and **`pnpm lint` is currently a dead command in both `chat` and `marketing`**, `eslint-config-next` is pinned two majors behind `next`, `@types/react` is a major behind React, and two deprecated conventions (`images.domains`, `middleware.ts`) still need migrating.

## Goal

All three apps run `next@16.3.0` with a green `build`, `test`, `tsc --noEmit`, and a **working** `lint`, verified on each app's live preview deployment.

Concretely, when this is done:

- `next` is `16.3.0` in `chat/package.json`, `marketing/package.json`, `api/package.json` (all three sit at `16.0.10` today).
- `pnpm lint` **runs and reports** in all three repos. Today it exits 1 in two of them without linting a single file:
[...]

PR #1930, opened 2026-08-04
fix(lint): restore a working `pnpm lint` on ESLint flat config

sweetmantech:
Part of #1929 (PR 2 of 8). Sequenced before the `next@16.3.0` bump so that the bump's "lint clean" claim actually means something.

## The bug

`pnpm lint` has been a dead command in this repo since the Next 16 upgrade:

```
$ pnpm lint
> next lint
Invalid project directory provided, no such directory: /…/chat/lint
ELIFECYCLE  Command failed with exit code 1
```

[`next lint` was removed in Next.js 16](https://nextjs.org/docs/app/guides/upgrading/version-16#next-lint-command), so `next` now parses `lint` as a positional project directory. **Not one file in this repo has been linted since.** `pnpm format` chains `&& pnpm lint`, so it has been failing at the end too.

## The fix

- `eslint` 8 → 9, `eslint-config-next` **14.2.13 → 16.3.0**. The lint config was two majors behind `next` itself.
- `.eslintrc.json` → `eslint.config.mjs`. Flat config is the Next 16 default, and ESLint 10 drops `.eslintrc` support entirely, so this had to happen either way.
- `lint` → `eslint .`, plus a `lint:fix` for the `--fix` behaviour `next lint --fix` used to provide.
- Dropped `eslint-plugin-storybook` and its `plugin:storybook/recommended` extend — there is no `.storybook/` directory and no `*.stories.*` file anywhere in the repo, so it was linting nothing.

## What turning it back on found

74 problems. They are two different things and this PR treats them differently.

**8 errors from rules the old config already enabled** (via `next/typescript`). Real debt that was only invisible because the command was broken — fixed here:

| File | Fix |
|---|---|
| `hooks/useArtists.tsx` | `menuVisibleArtistId` typed `any` → `string \| null`. `null` is the closed state; `Artist.tsx` now passes `artist.id ?? null` since the id is optional |
| `types/Agent.tsx` | `account_info: Array<any>` → the generated `Tables<"account_info">` row type |
| `hooks/usePWADownload.tsx` | `(navigator as any).standalone` → a narrow `Navigator & { standalone?: boolean }` |
| `tailwind.config.ts` | `require("tailwindcss-animate")` → a real import |
| `types/html2pdf.d.ts` | keeps `any`, with a justification. The package ships no types and returns a chained builder, so `any` is accurate rather than lazy |
| `hooks/useClickOutsideSelect.tsx` | **deleted** — zero importers anywhere (CLAUDE.md: "delete unused files/code that are not imported") |

**48 errors from `react-hooks/*`** — the React Compiler era of `eslint-plugin-react-hooks` that ships with `eslint-config-next` 16 and did not exist in 14.x. Nobody violated a rule that wasn't there. They span 46 files and each needs a genuine effect/ref refactor; reviewing 46 of those inside a tooling PR would be reckless. They report as **warnings** for now, with the reasoning written into the config, and get burned down separately. Nothing is silenced — they print on every run.

## Verification

| Gate | Before | After |
|---|---|---|
| `pnpm lint` | ❌ exit 1, **0 files linted** | ✅ exit 0 — 66 warnings, 0 errors |
| `pnpm test` | 88 files / 375 tests | ✅ 88 files / 375 tests, unchanged |
| `pnpm exec tsc --noEmit` | 7 errors | ✅ 7 errors, byte-identical set (all pre-existing, all in `lib/emails/__tests__/extractSendEmailResults.test.ts`) |

I did briefly regress the type check to 10 errors by narrowing `menuVisibleArtistId` to plain `string` — three call sites pass `null`/`undefined`. Caught it, corrected the type to `string | null`, and confirmed the set is back to the original 7.

`pnpm build` cannot complete on my machine — it fails collecting page data with `Missing Supabase credentials` and `ARWEAVE_KEY environment variable is missing`, which is my shell lacking prod secrets, not this diff. Compilation itself succeeds, and Tailwind's config is evaluated during compilation, so the `tailwindcss-animate` import change is exercised. Preview deployment will confirm the rest.

## Follow-up

The 48 `react-hooks` warnings want their own issue — several are `set-state-in-effect` in providers (`OrganizationProvider`), which is exactly the cascading-render pattern that rule exists to catch. Happy to file it.

---
## Summary by cubic
Restores a working `pnpm lint` after Next 16 removed `next lint` by migrating to ESLint flat config and updating lint tooling. Lint now runs via `eslint .` and exits 0 (63 warnings); tests are 85/356 after merging `main`; React Hooks rules warn for now.

- **Dependencies**
  - Upgrade `eslint` to v9 and `eslint-config-next` to v16.3.0.
  - Replace `.eslintrc.json` with `eslint.config.mjs` (flat config).
  - Remove `eslint-plugin-storybook` (no Storybook files present).
  - Update scripts: `lint` → `eslint .`; add `lint:fix`.
  - Regenerate `pnpm-lock.yaml` after merging `main` to resolve conflicts.

- **Bug Fixes**
  - Tighten types: `menuVisibleArtistId` to `string | null`; pass `artist.id ?? null`.
  - Narrow PWA standalone check with `Navigator & { standalone?: boolean }`.
  - Import `tailwindcss-animate` directly instead of `require`.
  - Use `Tables<"account_info">` for `account_info` arrays.
  - Keep `any` for `html2pdf.js` with justification in d.ts.
  - Remove unused `useClickOutsideSelect` hook.

<sup>Written for commit 8dbe1c51aa5854da2513c6d2fc2697ec669524c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sweetmantech:
## Preview verification: ✅ green (Chrome DevTools, logged in)

**Preview:** https://chat-irwov6yia-recoup.vercel.app
**Deployment sha:** `ec8fb9c9` — confirmed equal to this PR's head, not a stale build.

This PR is labelled "lint", but it changed two things with real UX surface, so I tested those specifically rather than just confirming the app boots.

---

### 1. `tailwindcss-animate`: `require()` → `import` — the risky one

This PR rewrote `tailwind.config.ts` from `plugins: [require("tailwindcss-animate")]` to a top-level `import`. If that broke plugin registration, **every animation in the app would silently stop** — the classes stay in the markup, but no CSS exists behind them. Nothing would throw, no test would fail, and it would ship.

Probed the live preview directly:

```js
const probe = document.createElement('div');
probe.className = 'animate-in fade-in zoom-in-95 slide-in-from-top-2 duration-300';
document.body.appendChild(probe);
getComputedStyle(probe)  // →
```

| Property | Value | |
|---|---|---|
| `animation-name` | `enter` | ✅ plugin keyframe registered |
| `animation-duration` | `0.3s` | ✅ `duration-300` resolves |
| `--tw-enter-opacity` | `0` | ✅ |
| `--tw-enter-scale` | `.95` | ✅ |
| `--tw-enter-translate-y` | `-.5rem` | ✅ |

The plugin is fully live. Tailwind generally is compiling too — the login modal renders with backdrop blur, shadow, and rounded corners:

Landing

---

### 2. `menuVisibleArtistId`: `""` → `null`

The other behaviour change. `useArtists` now initialises to `null` instead of `""`, and `Artist.tsx` passes `artist.id ?? null` because the id is optional. That state drives the artist dropdown via `isVisibleDropDown={artist.id === menuVisibleArtistId}`, so a mismatch would either strand a menu open or open several at once.

Worth noting for reviewers: the trigger is **`onContextMenu`** (right-click), not a visible button — which is why it does not appear in the accessibility tree. I drove it with a synthetic `contextmenu` event; the handler calls `preventDefault()`, so no native menu opens.

**Initial state — `null`, no menu open:**

Artists, menu closed

**Right-click "Ana Bárbara" → menu opens on that card only:**

Menu open on Ana Bárbara

**Right-click "Charlie Zaa" → menu *moves*; Ana Bárbara's closes:**

Menu moved to Charlie Zaa

Asserted programmatically rather than by eye, counting rendered `Remove` buttons across the whole grid:

| Step | `Remove` buttons in DOM | Open on | |
|---|---|---|---|
| Initial load | `0` | — | ✅ `null` renders no menu |
| Right-click Ana Bárbara | `1` | `Ana Bárbara` | ✅ |
| Right-click Charlie Zaa | `1` | `Charlie Zaa` | ✅ moved, not duplicated |
| Left-click a card (`setMenuVisibleArtistId(null)`) | `0` | — | ✅ closes |

Never more than one menu open, and the close path back to `null` works. Full state machine intact.

I did **not** click `Remove` — that deletes a real artist from the roster.

---

### 3. Auth + general health

Logged in as the account owner via Privy email OTP. Roster loads, sidebar renders, chat composer and model selector present:

Logged in

**Console:** clean. Logged out, the only two errors are the expected pre-auth `401` (`"Exactly one of x-api-key or Authorization must be provided"`). Logged in, a single benign warning: `"Attempted to log in, but user is already logged in."` No React, Tailwind, or type errors anywhere.

---

### Not covered

- The **48 `react-hooks` warnings** this PR downgrades to `warn` are unexercised by definition — they are latent correctness issues (several `set-state-in-effect` in `OrganizationProvider`), not behaviour this diff changes. Tracked as a follow-up on #1929.
- `usePWADownload`'s `navigator.standalone` narrowing only fires on iOS Safari standalone mode; not reachable in desktop Chrome. It is a type-only change with no runtime effect.
- Deleted `useClickOutsideSelect` had zero importers, so there is nothing to exercise.

Screenshots hosted on the `assets/pr1930` branch.

sweetmantech:
## Conflict resolved — `pnpm-lock.yaml` vs `main`

`main` moved ahead with #1934 (`ai` + `@ai-sdk/react` to the stable v6 line) and #1935 (reconnect via `WorkflowChatTransport`), which collided with this branch's lockfile.

**Regenerated the lockfile rather than hand-merging it.** Taking either side of a lockfile conflict leaves it inconsistent with the merged `package.json` — the manifest asks for one set of versions and the lock pins another, and it only surfaces later as a broken CI install. Took `main`'s copy as the base, re-resolved with `pnpm install`, then proved consistency with `pnpm install --frozen-lockfile`.

`package.json` auto-merged clean and keeps both sides:

| Package | Version | From |
|---|---|---|
| `eslint` | `^9.39.5` | this branch |
| `eslint-config-next` | `16.3.0` | this branch |
| `ai` | `6.0.241` | main |
| `@ai-sdk/react` | `3.0.243` | main |
| `eslint-plugin-storybook` | absent | this branch (removal preserved) |

`lint` script still `eslint .`.

### Re-verified after the merge

| Gate | Result |
|---|---|
| `pnpm lint` | ✅ exit 0 — 63 warnings, 0 errors |
| `pnpm test` | ✅ 85 files / 356 tests |
| `pnpm exec tsc --noEmit` | ✅ 7 errors, all still in `lib/emails/__tests__/extractSendEmailResults.test.ts` |

Two counts moved, and **both deltas come from `main`, not this branch**: warnings 66 → 63 and tests 88/375 → 85/356, because #1935 deleted the hand-rolled reconnect code and its tests. Flagging so the numbers in my earlier verification comment don't read as a regression here.

Issue #1928, opened 2026-08-04
Chat streams die at a ~120s per-connection cap: a long turn freezes the UI on a focused tab


sweetmantech:
Every reader connection to a chat workflow stream is capped at **~120 seconds**, regardless of how long the run lasts. The connection ends *cleanly* — a normal `data: [DONE]` with no `finish` chunk — so nothing downstream can tell it apart from a completed turn. This is the "~123 s ceiling" tracked as a row on [chat#1923](https://github.com/recoupable/chat/issues/1923), now measured directly. Business context, out of scope here: a user watching a long agent turn loses everything after the two-minute mark, including the final answer.

## Measured, not inferred

Headless run on `test-recoup-api` (`wrun_01KZ59E1HX5VV8FX2MJYXFQDY3`, chat `d626d367-296c-4b1a-aa9d-82227b8dc07c`), prompt = seven sequential 40 s sleeps (~280 s of work). Two consecutive readers, timed off the wire:

| Reader | Request | Ended | Last chunk | Run state |
|---|---|---|---|---|
| 1 | `GET /api/chat/{chatId}/stream` | **120 s** | `[DONE]` after 39 data lines | still running |
| 2 | `GET /api/chat/{chatId}/stream?startIndex=39` | **121 s** | `[DONE]` after 46 data lines | still running |

Reader 2 resumed correctly — first chunk was a continuation (`text-start` → *"After sleep 3…"*), no replay of the 39 chunks already seen. So **the resume contract works; the connection lifetime is the problem.**

## What this rules out

- **Not our route's `maxDuration`.** Both `app/api/chat/route.ts` and `app/api/chat/[chatId]/stream/route.ts` declare `maxDuration = 800`.
- **Not an abrupt kill or a network fault.** The stream terminates gracefully with `[DONE]`; a dropped socket would not.
- **Not an idle timeout.** Reader 1 carried chunks at 79 s and 119 s and still ended at 120 s.
- **Not caused by [api#808](https://github.com/recoupable/api/pull/808).** The decompose only made runs long enough to routinely cross two minutes.

Corroborating: `runAgentWorkflow.ts` already documents Workflow's runtime garbage-collecting a run's stream at **"observed ~2 m"**, which is the same number from the other direction.

## Why it matters more than it looks

[...]

PR #1935, opened 2026-08-04
feat(chat): reconnect via WorkflowChatTransport instead of hand-rolled recovery

sweetmantech:
Row 1 of the [chat#1923](https://github.com/recoupable/chat/issues/1923) plan — the fix for the UI stall.

Replaces our client-side stream-recovery machinery with **`@workflow/ai`'s `WorkflowChatTransport`**, which owns reconnection natively.

## Why the transport, and not one more fix

The transport is a **single generator**:

```js
for await (const chunk of postStream) { chunkIndex++; yield chunk; if (chunk.type === 'finish') gotFinish = true; }
if (!gotFinish) yield* this.reconnectToStreamIterator(options, runId, chunkIndex);
// …which loops: while (!gotFinish) { GET …/stream?startIndex=chunkIndex; … }
```

One continuous stream reaches `useChat`, so the turn settles exactly when `finish` arrives.

Our version tracked that same state across React refs — an in-flight flag, an edge-detected trigger, a probe schedule, a terminal-state branch — and **every fix exposed another unverified state**:

| Attempt | Failure |
|---|---|
| trigger on `useChat` status | fired once, never again |
| flag held across `await resumeStream()` | blocked recovery for the whole ~120 s connection |
| final stream end | fired nothing at all — 2 of 3 preview runs passed, the third hung after 9 recoveries |

There is no trigger to miss here, and no ref to hold.

## Deleted — 7 files

`createChunkCountingFetch`, `buildStreamReconnectUrl`, `shouldRecoverStalledStream`, `useStreamRecovery`, their tests, and the hook's wiring in `useVercelChat`. **`startIndex` is no longer ours to compute.**

## Kept, because the SDK cannot know it

- auth + `sessionId` / `chatId` via `prepareSendMessagesRequest`
- the reconnect URL repointed at `/api/chat/{chatId}/stream` — the SDK defaults to keying by **run id**
- a **204 override**: our api answers "nothing to resume" with a bodyless 204, which the SDK surfaces as `Failed to fetch chat: 204` and counts toward `maxConsecutiveErrors`. Mirrors upstream open-agents' wrapper.

## Version choice

`@workflow/ai@4.2.0` is **stable** and carries *newer* transport code than the 5.x beta upstream pins — I diffed them, and 4.2.0 has the mid-part-resume orphan filter (`createOrphanFilter`) that `5.0.0-beta.4` lacks. That matters here because we resume at arbitrary chunk indices.

Lockfile resolved with **default** resolution: **9 version changes**, `viem` and everything else unmoved. `--config.resolution-mode=lowest-direct` was tried and rejected — it *downgraded* `viem` 2.52.0 → 2.23.2 and broke the coinbase types. "Lowest" is not "least change".

## Checks

356 tests pass, `tsc --noEmit` clean, `next build` compiles.

Preview verification to follow in a comment. The gate is the one that has never held: **reconnects chain through every stream end, and the turn ends coherently — across 3 runs.**

---
## Summary by cubic
Switches chat reconnection to `@workflow/ai`’s `WorkflowChatTransport` to stop UI stalls on long runs by resuming streams until the `finish` chunk. Removes custom recovery code and aligns reconnects with our API.

- **Refactors**
  - Added `createWorkflowChatTransport` and switched `useChatTransport`; removed `useStreamRecovery` and related helpers/tests.
  - Reconnects call `/api/chat/{chatId}/stream` with auth; `startIndex` is computed by the transport and it streams until `finish`; treat `204` as “nothing to resume”.
  - Injected `sessionId`, `chatId`, and `recoupAccessToken` via `prepareSendMessagesRequest`.
  - Added `@workflow/ai@4.2.0`; lockfile updated.

<sup>Written for commit 602d22c24b12f5edd874208b983bfc089da7d917. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1935?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved chat connection handling with automatic stream resumption.
  * Added secure session and access-token handling for chat requests and reconnects.
  * Added a single retry for reconnects caused by expired authentication.
  * Gracefully handles completed streams during reconnection.

* **Bug Fixes**
  * Improved recovery from interrupted or incomplete chat responses.
  * Prevents unnecessary reconnect attempts when a stream has ended normally.


sweetmantech:
## Preview verification — 2 runs, 1 pass 1 fail. **Necessary but not sufficient.**

Tested after [api#810](https://github.com/recoupable/api/pull/810) merged (`d4109075`) and `test` was synced (`589c695e`), confirmed live: `access-control-expose-headers: x-workflow-run-id, x-workflow-stream-tail-index`.

### First: the header fix unblocked the transport

Before #810, the very first run died instantly with `Workflow run ID not found in "x-workflow-run-id" response header` and **3,804 chars** rendered (the model dropdown, nothing else). After: **131,306 chars** on run 1. The SDK reads that header unconditionally, so this PR was dead in the water without it.

### Run 1 — ✅ PASS

Streamed to **131,306 chars**, produced a complete answer, **sent the email**, and settled: composer returned to its send control and `isStreaming: false` agreed. Real content, correctly reported:

> *"The production API rejected the token with a 401, so per your instruction I retried against test-recoup-api… 3 artists, 4 connected socials — all real figures from the API"* — plus an unprompted caveat that the numbers came from the test environment.

### Run 2 — ❌ FAIL (stalled)

Heavier prompt (per-platform followers + recent post activity). Streamed strongly — **370k chars at 115 s, 555k at 230 s, 742k at 345 s**, well past the ~120 s cap — then froze.

| Observation | Value |
|---|---|
| Content frozen | 742,490 chars, unchanged across **100+ s** |
| `isStreaming` (server) | **false** — re-checked twice, not the iteration-boundary race |
| Composer | **stop control still showing** |
| Reconnect attempts | **zero** — no `GET …/stream` ever issued |

### The diagnosis, and why this PR cannot fix it alone

`WorkflowChatTransport` reconnects when the POST stream **ends without a `finish` chunk**. Here the POST stream **never ended at all** — so there was nothing for the transport to react to. It is holding an open connection to a run that finished server-side.

The client cannot act on a stream that never terminates. This is the api-side defect already tracked as row 2 on [chat#1923](https://github.com/recoupable/chat/issues/1923): `runAgentWorkflow` sends `sendStreamFinish` **only on the happy path**, has no `catch`, and its `finally` calls `closeChatStream` without a finish. Upstream open-agents does `sendFinish(writable).then(() => closeStream(writable))` on every exit path, commented *"so the chat is never permanently marked as streaming."*

**So the merge order is: api finish/close → this PR.** The transport is the right architecture — run 1 proves the whole path works end to end — but it needs the server to terminate its streams.

I am not claiming this fixes the stall on its own; run 2 says otherwise.

sweetmantech:
## Preview verification — both api prerequisites merged

Re-ran after [api#810](https://github.com/recoupable/api/pull/810) (`d4109075`) and [api#814](https://github.com/recoupable/api/pull/814) (`7a6f0094`) merged and api `test` was synced (`b3d2a203`). Confirmed `test` carries the finish fix before starting: a probe run returned `finish: 1 | DONE: 1 | terminal order: ['finish','DONE']`.

### Run 1 — ✅ PASS

Roster status prompt. Complete answer, **`Email sent ✅`** (send ID `3bccb8d9-701e-435d-b9ad-c8ebccf59450`), composer returned to its send control.

Content was real and self-aware: 3 artists / 4 connected profiles / 18,070 combined followers, with the model flagging that production had 401'd and it had fallen back to the test base as instructed, plus a scoping note that the roster covers only the credential's own account.

### Run 2 — ✅ PASS — **this is the prompt that previously hung**

Per-platform followers **plus** recent post activity — byte-for-byte the prompt that stalled at 742,490 chars on the unfixed api. Now: settled at **374 s**, complete answer, **`Email sent ✅`** (send ID `6cf28b95-e66c-4da3-8eb2-373c1ab5bbdb`).

It also reported an absence honestly rather than papering over it:

> *"there wasn't any to report. `GET /api/artists/{id}/posts` returned 0 posts for all three artists. Before settling for that, I triggered a fresh scrape … and polled twice over ~90 seconds … so \"0 posts\" is the accurate status, not a gap I papered over."*

### Run 3 — ⏳ in progress, verified alive

Heaviest prompt (scrape every connected social, summarise trends). Still running at **~30 minutes**, and confirmed alive server-side rather than assumed:

```
wrun_01KZ7RP1EAVZBHTQN4WJNXT   running   dur=1821s
```

It fanned out **53 scrape jobs** (all succeeded) and resolved **42 unique artists** across two org lists. Rendered text is static because no chunks are emitted while a long tool call executes — that is not a stall, and the workflow status is the evidence.

### Two corrections to my own method

1. **My "settled" detector was wrong.** It looked only for a square stop icon (`svg rect`); this UI also renders stop as an ✕ (`M18 6 6 18`). "No rect" never meant settled, so run 3 briefly looked finished when it was not. Runs 1 and 2 still stand — they are confirmed by a complete final answer and distinct email send IDs, not by the icon.
2. **`isStreaming` is not a usable oracle.** It read `false` while a turn streamed a further ~200,000 chars ([chat#1928](https://github.com/recoupable/chat/issues/1928#issuecomment-5185939156)). Liveness here is judged from the workflow run status instead.

### Status

The stall is fixed for both prompts that previously reproduced it. I will post run 3's terminal state when it lands rather than call this complete on 2 of 3.

sweetmantech:
## Run 3 — ❌ FAIL, but a **different** bug

Waited for the workflow to reach a terminal state rather than calling it early. It did:

```
wrun_01KZ7RP1EAVZBHTQN4WJNXT   completed   2026-08-05T02:06:23Z   (~36 min)
```

**15 minutes after the run completed, the client was still showing the stop control**, frozen at 7,806 chars, no final answer, no email. So this is a genuine UI stall.

### Root cause — not stream termination

```
Error: Failed to fetch chat: 401 {"status":"error","message":"Authentication token expired"}
```

The Privy access token has a **3,600 s TTL** (`exp - iat = 3600`). This turn ran **~36 minutes** and the session had already been open a while, so a reconnect fired with an expired token, got a `401`, and `WorkflowChatTransport` threw — leaving `useChat` in-flight forever.

Confirmed it is expiry and not a broken credential: the *currently* stored token is valid (`expiresIn 1838 s`) and the same authenticated call now returns **200**. A retry at the time would have succeeded.

### Why this is not the bug this PR fixes

The original stall was **the stream never terminating** — the api closed without a `finish` chunk, so the client had nothing to react to. That is fixed and demonstrated by run 2, which is byte-for-byte the prompt that previously hung and now completes in 374 s.

This is a distinct path: the stream terminates fine, but the **reconnect is rejected**, and a rejected reconnect has no terminal handling.

### The gap it exposes, in my code

`createWorkflowChatTransport` overrides `reconnectToStream` to swallow **204** (our "nothing to resume") but nothing else. A `401` propagates as a throw, and `useChat` is left in-flight with no error surfaced to the user — the same visible symptom, a different cause.

Two things worth doing, and I do not think either should silently ride along in this PR:

1. **Retry once on 401 with a freshly minted token.** `getAccessToken()` is called per request, so a refresh is available; the failure is that a stale token is used and never retried.
2. **Surface a terminal error when a reconnect cannot be recovered**, rather than hanging. Upstream's wrapper handles only 204 too — but upstream turns do not run for an hour, so they never hit this.

### Scoreboard

| Run | Prompt | Result |
|---|---|---|
| 1 | roster status | ✅ complete, email sent, settled |
| 2 | **the prompt that previously hung** | ✅ settled at 374 s, email sent |
| 3 | scrape every social, 36 min turn | ❌ stalled — **expired auth on reconnect**, not stream termination |

2 of 3, with the failure understood and attributable to a different defect.

sweetmantech:
## Rolled back the 401 retry (`4a4574ed`)

Reverted at Sweets' call — the branch is back to `602d22c2`, the pure transport swap, to stay closest to upstream open-agents.

**Why the retry may have been solving a phantom:** run 3 was ~36 minutes and **the laptop was shut partway through**. A suspended tab does not refresh its Privy token, so the 401 on reconnect is at least as well explained by sleep/resume as by an ordinary long-running turn. Fixing that in the transport would have been treating a symptom of tab suspension as a streaming defect.

Worth noting for later: on resume, a plain page refresh re-authenticates and the transcript loads correctly, so an auto-refresh (or a re-auth prompt) on wake is likely the right shape if this recurs — not a retry buried in the transport.

Also corrected from that discussion: the ephemeral sandbox key we mint (`DEFAULT_EPHEMERAL_KEY_TTL_MS`) is **60 minutes**, already raised from 15m in chat#1839. That is ours to lengthen, but it authenticates the **agent inside the sandbox**, not the browser's reconnect, so it is a separate knob from the one that failed here.

Re-running three verification runs, each kept under 60 minutes with the machine awake.

356 tests pass, tsc clean, lint clean.

sweetmantech:
## Preview verification round 2 — **3 / 3 passed**

Re-ran after rolling back the 401 retry (`4a4574ed`), so this tested `602d22c2` — the **pure transport swap**, no auth handling, closest to upstream open-agents. Machine kept awake throughout, each run inside the 1-hour Privy token window.

| Run | Prompt | Duration | Workflow | Client | Result |
|---|---|---|---|---|---|
| 1 | roster status | **525 s** (~9 min) | `completed` | settled, email sent | ✅ |
| 2 | **the prompt that originally hung** | **983 s** (~16 min) | `wrun_01KZ7YQZAJ7F3BWZ7DHQ4` `completed` | settled, email sent | ✅ |
| 3 | same, repeated for reproducibility | **826 s** (~14 min) | `wrun_01KZ7ZZEM3T8NVCVXNEYT` `completed` | settled, email sent | ✅ |

Every run is **6–8× past the ~120 s connection cap** ([chat#1928](https://github.com/recoupable/chat/issues/1928)) that used to freeze the UI, and in every case the client's terminal state matched the workflow's — no stalls, no truncation, no divergence.

### The case that proves it

Run 3 spent a long stretch on a `todo_write` tool call with a spinner and no visible progress — exactly the shape that used to mean "orphaned forever." It **resolved on its own with no page refresh and no intervention**: the workflow completed at `03:51:37`, and the composer returned to its send control on its own.

That distinction is the whole point of this work. Before, a spinner could mean *working* or *permanently hung*, and you could not tell without querying the server.

### Output quality held up

The agent reported gaps instead of padding them, unprompted:

> *"4 artists have no connected social profiles (DC The Don, MGMT, Cantora Records, Sean Price) — shown with zero platforms."*
> *"Several artists have 0 posts on record … stated plainly rather than estimated."*
> *"BennyJ504's posts endpoint threw a transient 500 on the first call; a retry succeeded (0 posts)."*

Run 3 tracked **2,028 posts across the roster** — heavier than either earlier run.

### On round 1's failure

Round 1 was 2/3: run 3 stalled with `Failed to fetch chat: 401 … Authentication token expired` after a ~36 minute turn **during which the laptop was shut**. A suspended tab does not refresh its Privy token, so that is at least as well explained by sleep/resume as by a long turn. The 401 retry added for it was **rolled back** rather than kept — tab suspension is not a streaming defect, and an auto-refresh on wake belongs at the app level, not buried in the transport.

Worth noting separately: the ephemeral sandbox key TTL (`DEFAULT_EPHEMERAL_KEY_TTL_MS`, currently 60 min, raised from 15 min in chat#1839) is ours to lengthen, but it authenticates the **agent inside the sandbox**, not the browser's reconnect — a different knob from the one that expired.

### Checks

356 tests pass, `tsc --noEmit` clean, lint clean, Vercel build passes.

Issue #1949, opened 2026-08-07
Report emails silently rejected at validation (59% of sends); stale run columns, dead IG handle, cold-load error toast


sweetmantech:
Tracking issue for a customer bug report received as a voice note on 2026-08-06. One customer session surfaced several compounding failures: monthly report emails that stopped arriving for ~9 months, a report computed against a social handle the artist had since changed, and a visible error in the app when she went in to fix it.

**Status: reproduced and root-caused on 2026-08-06.** Two of the four findings are systemic, not specific to this account. Reporting account is referenced by UUID only (`b1d40ce4-…`); identity intentionally omitted since this repo is public.

## Goal

1. A scheduled report either delivers or fails loudly. Never silently stops delivering while the task keeps running and the UI reports it as healthy.
2. `POST /api/emails` stops rejecting the majority of its traffic at validation.
3. When a connected social handle stops resolving, the system detects it rather than shipping a report full of empty numbers.
4. A cold page load does not show the user "An error occurred, please try again!" for a failure that is internal, expected, and recoverable.

## PRs (updated 2026-08-07)

Ordered by fix priority. The two page-load errors lead: they fire on **every** cold load of `/` for **every** account, they are the only findings a customer actually sees in the product, and they are what made this customer stop trying and reach for her phone.

| PR | Item | State |
|----|------|-------|
| [chat#1950](https://github.com/recoupable/chat/pull/1950) | **F4a: page-load error toast** — stream resume fires before auth + chat exist (401/404) | ✅ merged 2026-08-07 — see Done |
| api#TBD | F2a: record the rejection reason in `email_send_log` (diagnostic unblocker) | ⏳ not started |
| api#TBD | F2b: fix the dominant `POST /api/emails` rejection cause once visible | ⏳ not started |
| chat#TBD | F6: stop presenting advice as analysis when the data source is unavailable | ⏳ not started |
| chat#TBD | F5: create the schedule when a request specifies a cadence | ⏳ not started |
| api#TBD | F1: maintain `scheduled_actions.last_run` / `next_run` on every run | ⏳ not started |
| api#TBD | F3: never overwrite a good social record with an empty scrape result | ⏳ not started |
| api#TBD | F7: mirror scraped images to blob storage instead of storing expiring CDN URLs | ⏳ not started |

[...]

PR #1950, opened 2026-08-07
fix(chat): gate stream resume on auth and a real chat id

sweetmantech:
Implements item **F4a** of #1949 — the cold-load error toast.

## The bug

`useVercelChat` passed `resume: true` unconditionally to `useChat`, so **every mount** issued `GET /api/chat/{chatId}/stream?startIndex=0`. On a cold load of `/` that request fails two different ways:

| | Cause | Response |
|---|---|---|
| **401** | The GET beats Privy's session resolution, so it carries no `Authorization` header | `Exactly one of x-api-key or Authorization must be provided` |
| **404** | `NewChatBootstrap` mounts `<Chat>` with a client-generated placeholder UUID ([NewChatBootstrap.tsx L34](https://github.com/recoupable/chat/blob/main/components/VercelChat/NewChatBootstrap.tsx#L34)) that has no row behind it until provisioning resolves | `Chat not found` |

Either one hits `onError`, which raises `toast.error("An error occurred, please try again!")`. Reproduced 2/2 in #1949 on an account unrelated to the reporter, so it is global. To a non-technical customer it reads as "the home page is broken and I can't start a chat" — which is exactly how it was reported.

## The fix

Gate resume on **both** preconditions via a new `shouldResumeStream` helper:

- **Auth resolved** (`authenticated` from Privy), so the header is present.
- **A chat id that actually exists** — the route `chatId` param on `/sessions/[sessionId]/chats/[chatId]`, or `workflowChatId` once bootstrap resolves. Never the placeholder.

**This delays the resume, it does not skip it.** `useChat` re-runs its resume effect when the flag changes (`useEffect(() => { if (resume) chatRef.current.resumeStream() }, [resume, chatRef])` in `@ai-sdk/react@3.0.243`), verified against the installed package rather than assumed. So returning to a chat mid-turn on the canonical route still re-attaches once auth lands.

## Why a separate file

One exported function per file (repo principles). It also makes the four-way condition directly testable without mounting the whole chat tree.

## Tests

RED before GREEN. `lib/chat/__tests__/shouldResumeStream.test.ts` covers all five states:

| Scenario | Resume? |
|---|---|
| Cold home load, no auth, no chat | ✗ |
| Auth resolved, still a placeholder chat | ✗ |
| Real chat route, auth not yet resolved | ✗ |
| Real chat route, auth resolved | ✓ |
| Bootstrap minted a real chat id | ✓ |

```
Test Files  93 passed (93)
     Tests  397 passed (397)
```

`tsc --noEmit` clean for the changed files. ESLint on `hooks/useVercelChat.ts`: **4 warnings before, 3 after** — this removes one pre-existing warning and adds none. 0 errors.

## Not in scope

- **F4b** (`/api/accounts/undefined/credits` → 400) is the other half of the same cold load and ships as its own PR — different file, different race, no dependency between them.
- The `console.error` at the `onError` site violates the repo's no-production-logging rule but predates this change.

## Verification status

🔄 Preview verification pending — Done-when for F4a is *a cold load of `/` produces zero 4xx responses and no error toast*, which needs the deployed preview and a DevTools network trace. Results to follow as a comment on this PR.

---
## Summary by cubic
Fixes the cold-load error toast by gating stream resume on a resolved session and an existing route chat id. Implements F4a of chat#1949 and stops 401/404s and post-bootstrap resume loops, while keeping mid-turn re-attach working.

- **Bug Fixes**
  - Added `shouldResumeStream` and wired it into `useVercelChat` to resume only when `authenticated` and a real `routeChatId` exists (never on a bootstrapped chat).
  - Prevents early `GET /api/chat/{id}/stream` calls without `Authorization`, against placeholder ids, and the 204 resume loop after bootstrap.
  - Added unit tests covering five key states.

<sup>Written for commit 77b2c9ab63b4bf995542594ae18abeb3adaeb375. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1950?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/hooks/useVercelChat.ts
+++ b/hooks/useVercelChat.ts
@@ -219,14 +219,11 @@ export function useVercelChat({
       id,
       transport,
       // Re-attach to an in-progress response, so returning to a chat mid-turn
-      // keeps rendering instead of showing a frozen half-message. Gated until
-      // auth and a real chat id exist — unconditionally resuming 401s/404s on
-      // every cold load and surfaces as an error toast (chat#1949 F4a).
-      resume: shouldResumeStream({
-        authenticated,
-        routeChatId: chatId,
-        workflowChatId,
-      }),
+      // keeps rendering instead of showing a frozen half-message. Gated to an
+      // authenticated visit to an existing chat — resuming unconditionally
+      // 401s/404s on every cold load and surfaces as an error toast
+      // (chat#1949 F4a).
+      resume: shouldResumeStream({ authenticated, routeChatId: chatId }),
       experimental_throttle: 100,
       generateId: generateUUID,
       onError: (e) => {
--- a/hooks/useVercelChat.ts
+++ b/hooks/useVercelChat.ts
@@ -26,6 +26,7 @@ import getMimeFromPath from "@/lib/files/getMimeFromPath";
 import { getChatPath } from "@/lib/chat/getChatPath";
 import { useInitialMessageAutoSend } from "./useInitialMessageAutoSend";
 import { usePersistSelectedModel } from "./usePersistSelectedModel";
+import shouldResumeStream from "@/lib/chat/shouldResumeStream";
 
 interface UseVercelChatProps {
   id: string;
@@ -217,9 +218,15 @@ export function useVercelChat({
     useChat({
       id,
       transport,
-      // Re-attach to an in-progress response on mount, so returning to a chat
-      // mid-turn keeps rendering instead of showing a frozen half-message.
-      resume: true,
+      // Re-attach to an in-progress response, so returning to a chat mid-turn
+      // keeps rendering instead of showing a frozen half-message. Gated until
+      // auth and a real chat id exist — unconditionally resuming 401s/404s on
+      // every cold load and surfaces as an error toast (chat#1949 F4a).
+      resume: shouldResumeStream({
+        authenticated,
+        routeChatId: chatId,
+        workflowChatId,
+      }),
       experimental_throttle: 100,
       generateId: generateUUID,
       onError: (e) => {
--- a/lib/chat/__tests__/shouldResumeStream.test.ts
+++ b/lib/chat/__tests__/shouldResumeStream.test.ts
@@ -7,7 +7,6 @@ describe("shouldResumeStream", () => {
       shouldResumeStream({
         authenticated: false,
         routeChatId: undefined,
-        workflowChatId: undefined,
       }),
     ).toBe(false);
   });
@@ -19,7 +18,21 @@ describe("shouldResumeStream", () => {
       shouldResumeStream({
         authenticated: true,
         routeChatId: undefined,
-        workflowChatId: undefined,
+      }),
+    ).toBe(false);
+  });
+
+  it("does not resume a freshly bootstrapped chat", () => {
+    // Verified on the preview deployment: gating on the bootstrap-minted id
+    // made `resume` flip false -> true once provisioning resolved, and
+    // `WorkflowChatTransport`'s `while (!gotFinish)` loop then reissued the
+    // resume GET 118 times against a brand-new chat that returns 204. A chat
+    // created seconds ago has no in-flight turn to re-attach to, so the
+    // bootstrap id is deliberately not an input to this decision.
+    expect(
+      shouldResumeStream({
+        authenticated: true,
+        routeChatId: undefined,
       }),
     ).toBe(false);
   });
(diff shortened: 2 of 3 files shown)


sweetmantech:
## Preview verification — and a regression this caught

Tested with Chrome DevTools against the preview for `77b2c9ab`, cross-checked against production **on the same account, in the same browser, in the same session**, so the before/after is like-for-like.

**Headline: the first version of this fix was wrong, and preview testing is the only thing that found it.** Details in "The regression" below. The current commit is verified clean.

| Environment | Deployment |
|---|---|
| Preview | `chat-kccmxz2a1-recoup.vercel.app` (built from `77b2c9ab`, confirmed via the deployment's sha) |
| Production | `chat.recoupable.dev` |
| Account | `fb678396-a68f-4294-ae50-b8cacf9ce77b` |

Preview runs a different Privy app (`cmc52us2g01rtjr0m0k2m5fkv` vs prod `cm4t193tx03w553u94ib9ucnh`) and points at `test-recoup-api.vercel.app`, so this required a separate sign-in.

---

## 1. Production baseline — both F4 bugs reproduce

Reproduced on every cold load. One representative trace:

| reqid | Request | Documented in #1949 | Actual |
|---|---|---|---|
| 1100 | `GET /api/chat/8f9e070b-…/stream?startIndex=0` | 401 or 404 | **404** |
| 1116 | `GET /api/accounts/undefined/credits` | 400 | **400** |
| 1153 | `GET /api/accounts/undefined/credits` | 400 | **400** |

**Three 4xx on a single cold load.** The `undefined` credits request fired twice that run. The error toast appeared on **5 of 5** loads, confirmed in the accessibility tree:

```
region "Notifications Alt+T"
  alert  StaticText "An error occurred, please try again!"
```

Production home, cold load

> **On the toast screenshot:** I could not photograph the toast itself. It lives ~2 seconds and its container renders at `y=617`, exactly the viewport's bottom edge, so it is present in the a11y tree but below the fold. Four capture approaches failed (immediate capture, element capture, cloning the node, neutralising the auto-close timer). The DOM assertion above is the evidence for the toast; the screenshot is the page state only. Flagging rather than implying a photo exists.

---

## 2. The regression this caught

The first pass gated resume on `authenticated && (routeChatId ?? workflowChatId)`. Unit tests passed. The preview said otherwise:

| Load | `/stream` requests |
|---|---|
| Production (home) | 1 |
| **Preview, first attempt (home)** | **118** |

Including the bootstrap id meant `resume` flipped `false → true` the moment provisioning resolved. `useChat` re-runs its resume effect on that change, and `WorkflowChatTransport`'s `while (!gotFinish)` loop then hammered a chat created seconds earlier — one with no stream to re-attach to, answering `204` every time.

Every response was a 204, so **a naive "zero 4xx" check would have passed** while the change was materially worse than the bug it replaced.

Fixed in `77b2c9ab` by dropping the bootstrap id from the decision entirely. A resume is only ever for returning to an existing chat that may still be mid-turn; a freshly minted chat is never that. `useMessageLoader` already draws this exact line for the same reason.

---

## 3. Preview after the fix — Done-when verified

`#1949` F4 Done-when: *a cold load of `/` produces zero 4xx responses and no error toast.*

| Run | Signed in | Resources | `/stream` | `/accounts/undefined/credits` | Toasts |
|---|---|---|---|---|---|
| Signed-out cold load | no | 87 | **0** | **0** | **0** |
| Cold load 1 | yes | 130 | **0** | **0** | **0** |
| Cold load 2 | yes | 130 | **0** | **0** | **0** |
| Cold load 3 | yes | 131 | **0** | **0** | **0** |

Each authenticated run used an 8–12s settle window, deliberately past the point where the 118-request storm began, so a passing result is meaningful rather than measured too early.

Full network audit of one authenticated load: **37 requests, every one 200/304, zero 4xx.** Credits resolved to a real UUID (`GET /api/accounts/fb678396-…/credits [200]`) rather than `undefined`.

Preview home, authenticated cold load

Preview home, signed out

---

## 4. Regression check — resume still works

Opened an existing chat on the canonical route (`/sessions/93e04bcc-…/chats/d81b87ba-…`, owned by the signed-in account) on **both** environments:

| Environment | `/stream` requests | Messages render |
|---|---|---|
| Production | 16, 19 | yes |
| Preview | 37, 24 | yes |

Resume still fires on the canonical route — the gate opens, which was the risk with an auth condition.

**One thing to flag, unrelated to this PR.** Both environments issue *many* resume requests on an existing completed chat, because the transport's `while (!gotFinish)` loop keeps reissuing against a chat with no active stream. Prod does it too, at overlapping magnitude, so **this is pre-existing and not introduced here** — but 16–37 requests to open one chat is worth its own issue. This PR does not make it better or worse; it only removes the home-page case.

A first attempt used a chat owned by a *different* account and returned `403` on the stream and session reads. That is correct authorization behaviour, not a defect, but I re-ran against an owned chat so the result could not be dismissed as a 403 path.

---

## Summary

| Done-when criterion | Result |
|---|---|
| Cold load of `/` produces zero 4xx | ✅ 3/3 authenticated + 1 signed-out |
| No error toast | ✅ 0 toasts, 4/4 loads (5/5 on prod for contrast) |
| Existing-chat resume not broken | ✅ fires on the canonical route, messages render |
| Production still reproduces the bug | ✅ 3 × 4xx and a toast on one load |

`397 tests`, `tsc --noEmit` clean, ESLint 0 errors (4 warnings before this change on `useVercelChat.ts`, 3 after).

**F4b is not covered here.** `/api/accounts/undefined/credits` did not fire on these preview runs, but that is a timing-dependent race and this branch contains no fix for it — its absence is not evidence. That is #1951's to prove.

PR #1993, opened 2026-08-21
feat(music): Music nav item and the /music gallery

sweetmantech:
The `/music` entry point: a **Music** item in the left nav and a page that lists everything the account has generated. First of two chat PRs for recoupable/chat#1992.

> **Depends on recoupable/api#849** (`GET /api/music`). Until that ships the page renders its empty state rather than breaking, but it has nothing to list. **Review order: recoupable/docs#308 → recoupable/database#60 → recoupable/api#848 → recoupable/api#849 → this.**

Designs approved 2026-08-21 and embedded in the issue.

## What a user gets

Click **Music** in the rail (or the mobile drawer) → `/music` → every song they have generated, newest first, with a status pill, an inline player, and a per-card download. In-flight generations show as Queued or Processing and the list refreshes itself until they land.

The generate form is deliberately **not** in this PR; it is the second one. This PR is still a complete slice on its own: the nav item, the route, and the gallery are all user-visible and testable.

## Notable details

**Polling stops when nothing is rendering.** `refetchInterval` is a function of the query data: 5s while any generation is `pending` or `processing`, `false` otherwise. A gallery of finished songs is static, and a fixed interval would bill every idle tab a request every five seconds forever.

**The nav item is wired in both places.** `Sidebar/Menu.tsx` and `SideMenu/SideMenu.tsx` are separate components that each re-import the nav items and compute their own active flags. Adding to only one is how a nav item ends up missing on mobile.

**A native `<audio controls>`, not a custom player.** It carries play, seek, and volume, is keyboard accessible, and needs no state from us. Rebuilding it would be a lot of surface for no gain.

**Status pills get a stronger tint in dark mode.** The light-mode 8% wash specified in DESIGN.md is effectively invisible on the near-black card, so dark uses 20% with the brighter hue.

**Empty-state copy is org-aware**, using `selectedOrgId` from `useOrganization` — "no songs in this organization yet" reads very differently from "no songs yet" when you have just switched context.

## Verification

- `pnpm exec tsc --noEmit` — clean, zero errors in the whole repo.
- `pnpm exec eslint` on every touched path — clean.
- **Tests**: the existing `SecondaryNav` test gains a Music case (RED first: it failed on the missing `view music` button before the component existed), plus 4 new tests on `MusicGenerationCard` covering the download link on a completed song, its absence while generating, the verbatim failure message, and the prompt-as-title fallback.
[...]

Issue #1992, opened 2026-08-21
Music generation end-to-end: /music page + POST/GET /api/music (MiniMax Music 3 via fal)


sweetmantech:
Tracking issue for the first end-to-end music-generation slice of the label-services platform: a `Music` item in the left nav → a `/music` page (fal-style generate UI + gallery of past generations), backed by new `POST/GET /api/music` endpoints that run [MiniMax Music 3](https://fal.ai/models/minimax/music-3) on fal.ai, a new `music_generations` table, and processing-status visibility while a song renders. Nothing exists today — the api repo has fal image/video endpoints ([`lib/fal/server.ts`](https://github.com/recoupable/api/blob/main/lib/fal/server.ts), `lib/content/video/`) but no audio generation, no `fal.queue` usage, and no music tables (grepped all 182 migrations).

Business context (out of scope here): this is one brick of "end-to-end record label services" — generated songs will later feed catalogs/releases; none of that downstream logic is in this issue.

> **Decision 2026-08-21 (@sweetmantech): v1 has no artist selector.** Songs are scoped to a personal account or an organization only — the artist axis (`artist_account_id` on the row, the roster picker in the form, artist-filtered reads) is dropped from this issue on a KISS call. Artist attribution moves to Phase 2 below. Everything downstream of that — the contract, the schema, the UI spec, and the ownership architecture decision — is written to personal/org scope.

> **Decision 2026-08-21 (@sweetmantech): the table is 13 columns, not 24.** On review of database#60, anything another system already knows came out of `music_generations`: generation parameters ride along as durable workflow arguments, the resolved seed and the step timeline are readable through `fal_request_id` and `workflow_run_id`, and credits are accounted in `usage_events`. Dropped: `organization_id`, `requested_duration_seconds`, `seed`, `num_inference_steps`, `guidance_scale`, `credits_charged`, `mime_type`, `file_size_bytes`, `source_url`, `title`, `logs`. This reverses two decisions recorded below (the row carrying its own `logs`, and a separate org column) and **invalidates part of the already-merged contract in docs#308**, which is tracked as its own row below.

[...]

PR #1994, opened 2026-08-21
feat(music): generate form with full fal parameter parity

sweetmantech:
The generate form on `/music`: prompt, lyrics, and the full fal parameter set behind an Additional Settings disclosure. Second of two chat PRs for recoupable/chat#1992, and the last row of its PR matrix.

> **Stacked on recoupable/chat#1993** (the nav item and gallery), so its files appear in this diff until it merges. **Full order: recoupable/docs#308 → recoupable/database#60 → recoupable/api#848 → recoupable/api#849 → recoupable/chat#1993 → this.** Also depends on `POST /api/music` from recoupable/api#848.

Designs approved 2026-08-21 and embedded in the issue.

## What a user gets

Type a prompt and lyrics, hit **Generate**, and the new song appears in the gallery below as Queued, then Processing, then playable — without leaving the page. The gallery's own conditional polling (shipped in #1993) does the watching; this PR just invalidates the query on success so the pending row shows up immediately rather than up to five seconds later.

## Parameter parity with fal, deliberately

Additional Settings exposes **all four** parameters the fal playground does — duration, seed, num inference steps, guidance scale — with fal's own ranges and defaults, read off the live form rather than from memory: duration 1–300 (we floor at 10), steps 1–100 default 30, guidance 0–20 step 0.5 default 1.7, seed free text defaulting to random.

The earlier draft hid steps and guidance on the grounds that they are flow-matching internals. That was the wrong call and was reversed on review: hiding two of four means we own a judgment about which knobs matter, and we would re-own it on every model swap. Full parity makes the UI a thin, predictable surface over the model.

Each control carries fal's affordances too: an info tooltip per input (the model's own parameter description), a reset-to-default eraser on each slider, and a randomize button on seed. That asymmetry is fal's, not an inconsistency of ours.

## Notable details

**A native `input[type=range]`, not a Radix slider.** `@radix-ui/react-slider` is not a dependency here, and adding one for a single control would not earn its weight. The native control is keyboard and screen-reader accessible as-is.

**Price is quoted in dollars with credits secondary** (`Costs $0.30 for 60s (30 credits)`) and reprices live as the duration slider moves. This restates the API's pricing constants client-side, which is a deliberate, commented duplication: quoting a price cannot cost a round trip per slider drag, and the API stays authoritative by freezing the real price onto the row at creation.

**API error messages surface verbatim.** `insufficient_credits` becomes a specific toast; a rejected lyrics field shows what the API said. A generic "something went wrong" would hide the two failures a user can actually act on.

**Blank seed is omitted, not sent as null** — the API treats an absent seed as "choose one" and reports back the seed it used.

## Verification

- **8 new tests** on the form: the disabled-until-filled rule, the dollar quote, the exact payload including the omitted blank seed, the collapsed-by-default disclosure, repricing on a slider move ($0.30 → $0.60), the randomize button filling the field, a slider reset returning to its documented default, and Generate disabled while one is in flight. Plus 4 on the shared pricing helpers asserting the client quote matches what the API charges (30 credits at 60s, the 15-credit floor, 150 at 300s).
- `pnpm exec tsc --noEmit` — clean, zero errors in the whole repo.
- `pnpm exec eslint` on every touched path — clean.
- `pnpm build` — the page compiles and is emitted (`.next/server/app/music/page.js`). The full build cannot complete in this checkout: with no local `.env`, `next build` clears compilation and then fails collecting page data for `/api/agent-creator`, `/api/agent-templates`, and `/api/upload`, three routes this PR does not touch that construct clients at module load. Supplying dummy Supabase vars just moved the failure to the next unrelated route, which is the confirmation that it is environmental. Build verification proper happens on the Vercel preview.

**No preview screenshots yet**, and no end-to-end run: `POST /api/music` does not exist until recoupable/api#848 is live, which itself waits on the migration. Once the chain lands I will run a real generation through the preview and post desktop and mobile screenshots plus the observed state transitions, per the house rule for frontend PRs.

Implements the chat(form) row of the PR matrix in recoupable/chat#1992 — the final row.

---
## Summary by cubic
Adds the /music generate form with fal parameter parity and correct org scoping. Previously no on-page generation and org scope used organization_id (ignored); now users generate in place, blank seed is omitted, price quotes live with duration, API errors surface verbatim, the gallery shows rows immediately, and org scope is sent as account_id so items land in the selected org. Guidance slider now uses a 0.1 step so the 1.7 default is reachable.

- Uses fal ranges/defaults with one intentional deviation: guidance scale step 0.1 (fal’s 0.5 made 1.7 unreachable on a range input). Native range inputs instead of `@radix-ui/react-slider`. Pricing is duplicated client-side for responsiveness; the API freezes the authoritative price at creation.
- Verify payload mapping in `useCreateMusicGeneration` and `lib/music/createMusicGeneration.ts`: blank seed omitted; org scope passed as account_id; parameter names match the API (`num_inference_steps`, `guidance_scale`).
- Tests cover disabled-until-filled, payload (including omitted seed), live repricing, reset/randomize controls, collapsed-by-default settings, in-flight disable, pricing parity, and that slider defaults sit on their step grid.

**Rollout**
- Requires POST /api/music and gallery polling via GET /api/music. No client migrations.

<sup>Written for commit 08e83850ec16b41215d6a2253f71cd250ef6b9ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1994?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/components/MusicPage/MusicAdditionalSettings.tsx
+++ b/components/MusicPage/MusicAdditionalSettings.tsx
@@ -1,0 +1,114 @@
+"use client";
+
+import { ChevronDown, Shuffle } from "lucide-react";
+import { cn } from "@/lib/utils";
+import { Input } from "@/components/ui/input";
+import MusicFieldLabel from "./MusicFieldLabel";
+import MusicSliderField from "./MusicSliderField";
+import { MUSIC_DEFAULTS, MUSIC_HINTS, MUSIC_RANGES } from "@/lib/music/const";
+
+export interface MusicSettings {
+  duration: number;
+  seed: string;
+  numInferenceSteps: number;
+  guidanceScale: number;
+}
+
+export interface MusicAdditionalSettingsProps {
+  isOpen: boolean;
+  onToggle: () => void;
+  settings: MusicSettings;
+  onChange: (settings: MusicSettings) => void;
+}
+
+const MusicAdditionalSettings = ({
+  isOpen,
+  onToggle,
+  settings,
+  onChange,
+}: MusicAdditionalSettingsProps) => {
+  const set = <K extends keyof MusicSettings>(key: K, value: MusicSettings[K]) =>
+    onChange({ ...settings, [key]: value });
+
+  return (
+    <div className="border-t pt-3">
+      <button
+        type="button"
+        onClick={onToggle}
+        aria-expanded={isOpen}
+        className="flex w-full items-center justify-between gap-2"
+      >
+        <span className="text-sm font-medium">Additional Settings</span>
+        <span className="inline-flex items-center gap-1 text-sm text-muted-foreground">
+          {isOpen ? "Less" : "More"}
+          <ChevronDown
+            className={cn("size-4 transition-transform", isOpen && "rotate-180")}
+          />
+        </span>
+      </button>
+
+      {isOpen && (
+        <div className="mt-3 grid gap-3 sm:grid-cols-2">
+          <MusicSliderField
+            id="music-duration"
+            label="Duration"
+            hint={MUSIC_HINTS.duration}
+            {...MUSIC_RANGES.duration}
+            value={settings.duration}
+            defaultValue={MUSIC_DEFAULTS.duration}
+            format={value => `${value}s`}
+            onChange={value => set("duration", value)}
+          />
+
+          <div className="flex flex-col gap-2 min-w-0">
+            <MusicFieldLabel htmlFor="music-seed" label="Seed" hint={MUSIC_HINTS.seed} />
+            <div className="flex items-center gap-2">
+              <Input
+                id="music-seed"
+                inputMode="numeric"
+                placeholder="random"
+                value={settings.seed}
+                onChange={event =>
+                  set("seed", event.target.value.replace(/[^0-9]/g, ""))
+                }
+                className="min-w-0"
+              />
+              <button
+                type="button"
+                aria-label="Randomize seed"
+                onClick={() =>
+                  set("seed", String(Math.floor(Math.random() * 2_147_483_647)))
+                }
+                className="shrink-0 inline-flex items-center justify-center size-9 rounded-xl border hover:bg-muted transition-colors"
+              >
+                <Shuffle className="size-4" />
+              </button>
+            </div>
+          </div>
+
+          <MusicSliderField
+            id="music-steps"
+            label="Num Inference Steps"
+            hint={MUSIC_HINTS.numInferenceSteps}
+            {...MUSIC_RANGES.numInferenceSteps}
+            value={settings.numInferenceSteps}
+            defaultValue={MUSIC_DEFAULTS.numInferenceSteps}
+            onChange={value => set("numInferenceSteps", value)}
+          />
+
+          <MusicSliderField
+            id="music-guidance"
+            label="Guidance Scale"
+            hint={MUSIC_HINTS.guidanceScale}
+            {...MUSIC_RANGES.guidanceScale}
+            value={settings.guidanceScale}
+            defaultValue={MUSIC_DEFAULTS.guidanceScale}
+            onChange={value => set("guidanceScale", value)}
+          />
+        </div>
+      )}
+    </div>
+  );
+};
+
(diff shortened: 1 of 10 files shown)


sweetmantech:
## Conflicts resolved, and preview verified end to end

Signed in as `sweetman@recoupable.com` on `chat-git-feat-music-generate-form-recoup.vercel.app`, against the synced `test-recoup-api`. **A real song generated through this form.**

### The conflicts

Same class as recoupable/api#849: this branch was cut from the gallery branch before its rework, so it collided with `main` on all four files chat#1993 later touched. Rebuilding on `main` with only the form work was cleaner than resolving four conflicts and then reworking the result. **25 files down to 10**, now `MERGEABLE`.

### The substantive fix: `organization_id`

The form sent `organization_id` for an org-scoped generation. The API **ignores that field outright**, so an org song would have landed as personal with no error — a wrong answer rather than a failure, and invisible in testing unless you checked ownership. Organizations are accounts, so scope is now sent as `account_id: selectedOrgId`, riding the same override path the rest of the app uses.

### End to end, through the form

| Step | Result | |
|---|---|---|
| Generate disabled until prompt **and** lyrics | disabled with neither, still disabled with prompt only, enabled with both | ✅ |
| Submit | button became a disabled **Generating** | ✅ |
| New generation appears | card rendered **Queued** above the existing song, "2 songs" | ✅ |
| Reaches terminal state | **Completed in ~30s, with no reload** — the conditional poll | ✅ |
| Result is real | new card carries a player and download pointing at `music/b85e5621-….wav` | ✅ |
| Generate re-enabled after success | yes | ✅ |

### Controls

| Check | Result | |
|---|---|---|
| Additional Settings collapsed by default | yes, expands to all four params | ✅ |
| Ranges | duration 10–300 step 5, steps 1–100, guidance 0–20 | ✅ |
| Info tooltips | six: Prompt, Lyrics, Duration, Seed, Num Inference Steps, Guidance Scale | ✅ |
| Reset buttons | three, each returning its slider to the documented default | ✅ |
| Randomize seed | fills the field with a real seed | ✅ |
| Cost repricing | 20s → `$0.15 (15 credits)`, 120s → `$0.60 (60)`, 300s → `$1.50 (150)` | ✅ |
| Credit floor | 20s stays at 15 credits, matching `max(15, ceil(d × 0.5))` | ✅ |
| Mobile 390px | form 342px, all four fields and the randomize button inside the viewport, no page overflow | ✅ |

### A third bug, found on the preview

**The Guidance Scale slider could not reach its own default.** The readout said **1.7** while the thumb sat at **1.5**: a range input snaps to its step grid, and 1.7 is not a multiple of the 0.5 step it carried. So the control contradicted its own label, and the default became unreachable the moment anyone dragged it. fal can use 0.5 because its control is a number field you can type into; a range cannot.

Now step 0.1, verified on the deployed build: thumb at 1.7, drag to 12, reset returns to exactly 1.7. Added a test asserting all three sliders' defaults sit on their own step grid, and **confirmed it fails when the step is put back to 0.5** — a real guard, not a decorative one.

### Screenshots

Desktop, form with Additional Settings expanded:



Immediately after generating, with the filled form and the new song below:



Mobile at 390px:



### One note on method

Mid-session an OTP-entry script of mine typed digits into the page's own range inputs, leaving Num Inference Steps and Guidance Scale both reading **8**. That is test pollution, not a product bug — both reset cleanly to 30 and 1.7, which is what the "after generate" screenshot was taken before I corrected. Calling it out so the stray values in that image are not read as a defect.

### Local checks

`components/MusicPage` + `components/Sidebar`: **29/29**. `tsc --noEmit` clean across the repo. `eslint` clean.

This is the last row of the PR matrix on recoupable/chat#1992.

PR #1995, opened 2026-08-24
fix(music): make the download button actually download

sweetmantech:
Reported in use: clicking the download button on a completed song **navigated the current tab to the audio file** instead of saving it, dumping the user out of `/music` onto a bare audio player.

## Cause

The card rendered `<a href={audio_url} download>`, and **the `download` attribute is ignored for cross-origin URLs**. Our audio is served from Supabase Storage (`…supabase.co`) while the app runs on its own domain, so the browser drops the attribute and treats the link as ordinary navigation. No amount of markup fixes this.

## Fix

Supabase Storage honours a `download` query parameter by returning `Content-Disposition: attachment`, which forces a real download regardless of origin — so the fix belongs in the URL, not the markup. Verified against a live object rather than taken from docs:

```
plain URL                  → content-type: audio/wav
?download=my-song.wav      → content-disposition: attachment; filename=my-song.wav
```

Because a real download is achievable, the `target="_blank"` fallback the reporter offered is not needed — the tab is never navigated at all.

Files also now save under a name slugified from the prompt (`genre-dream-pop-bpm-104.wav`) rather than the generation uuid, keeping whatever extension the stored object actually has so a future model returning mp3 still saves correctly.

## The test that let this ship

The original assertion was:

```ts
expect(download.hasAttribute("download")).toBe(true);
```

Which passed, and always would have — the attribute *was* present. It just does nothing cross-origin. That assertion is replaced with ones about what the URL actually does, so the same class of bug cannot pass again:

```ts
expect(href).toContain("download=");
expect(href).toContain("download=genre-acoustic-pop.wav");
```

## Verification

- **9 new tests** on `musicDownloadUrl` / `musicDownloadFilename`, each RED before GREEN: attachment param, appending to a URL that already has a query, filename encoding, extension preserved from the object, long-prompt truncation, the no-usable-characters fallback, and the null case.
- **3 card tests** covering the rendered href.
- Full chat suite: **471 tests / 105 files pass**.
- `tsc --noEmit` and `eslint` clean.

Preview verification to follow in a comment: clicking download saves the file and leaves `/music` on screen.

Fixes the download row of the PR matrix in recoupable/chat#1992.

---
## Summary by cubic
Make the Music “Download” button trigger a real file download and keep users on /music. Previously, clicking “Download” navigated to the audio file (browsers ignored the cross‑origin download attribute); now we append a Supabase `download` query to force `Content-Disposition: attachment`. Files also save under a readable, prompt‑derived name with the correct extension.

- Updates the card to build `href` with `musicDownloadUrl(audioUrl, musicDownloadFilename(prompt, audioUrl))`; still hides the link when not completed.
- Adds `lib/music/musicDownloadUrl.ts`:
  - `musicDownloadUrl` appends `?download=<encoded>` (or `&download=`) to the storage URL.
  - `musicDownloadFilename` slugifies the prompt, preserves the object’s extension, truncates to 64 chars, and falls back to `generated-song.<ext>`.
- Replaces brittle tests that only checked for a `download` attribute with assertions on the generated URL and filename; adds unit tests for edge cases (existing query params, encoding, long/empty prompts).

<sup>Written for commit 896844325d0d755d8097b63f3c4eb5b74e5826a9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1995?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Music downloads now use readable filenames based on the generation prompt.
  * Downloaded audio files use the correct file extension, defaulting to WAV when unavailable.
  * Download links now include the generated filename for a clearer download experience.

* **Bug Fixes**
  * Improved handling of missing or invalid audio URLs when preparing downloads.

### DIFF
--- a/components/MusicPage/MusicGenerationCard.tsx
+++ b/components/MusicPage/MusicGenerationCard.tsx
@@ -3,6 +3,7 @@
 import { Download } from "lucide-react";
 import MusicStatusPill from "./MusicStatusPill";
 import { formatDuration } from "@/lib/music/formatDuration";
+import { musicDownloadFilename, musicDownloadUrl } from "@/lib/music/musicDownloadUrl";
 import type { MusicGeneration } from "@/types/Music";
 
 const MusicGenerationCard = ({ generation }: { generation: MusicGeneration }) => {
@@ -10,6 +11,12 @@ const MusicGenerationCard = ({ generation }: { generation: MusicGeneration }) =>
   // The API returns no title: the prompt is what the user wrote and what
   // identifies the song to them.
   const title = generation.prompt;
+  const downloadHref = generation.audio_url
+    ? musicDownloadUrl(
+        generation.audio_url,
+        musicDownloadFilename(generation.prompt, generation.audio_url),
+      )
+    : null;
 
   return (
     // min-w-0 because a grid item defaults to min-width:auto and will not
@@ -48,7 +55,7 @@ const MusicGenerationCard = ({ generation }: { generation: MusicGeneration }) =>
             <track kind="captions" />
           </audio>
           <a
-            href={generation.audio_url ?? undefined}
+            href={downloadHref ?? undefined}
             download
             aria-label={`Download ${title}`}
             className="shrink-0 inline-flex items-center justify-center size-9 rounded-xl border hover:bg-muted transition-colors"
--- a/components/MusicPage/__tests__/MusicGenerationCard.test.tsx
+++ b/components/MusicPage/__tests__/MusicGenerationCard.test.tsx
@@ -24,8 +24,32 @@ describe("MusicGenerationCard", () => {
     render(<MusicGenerationCard generation={generation()} />);
 
     const download = screen.getByRole("link", { name: /download genre: acoustic pop/i });
-    expect(download.getAttribute("href")).toBe("https://cdn.example/music/abc.wav");
     expect(download.hasAttribute("download")).toBe(true);
+  });
+
+  // The download attribute alone shipped broken: browsers ignore it
+  // cross-origin, and our audio is on Supabase while the app is not, so the
+  // link navigated the tab to a bare audio file. Asserting the attribute
+  // exists is what let that through, so assert the url does the work.
+  it("asks storage for an attachment so the download survives being cross-origin", () => {
+    render(<MusicGenerationCard generation={generation()} />);
+
+    const href = screen
+      .getByRole("link", { name: /download genre: acoustic pop/i })
+      .getAttribute("href") as string;
+
+    expect(href).toContain("download=");
+    expect(href.startsWith("https://cdn.example/music/abc.wav")).toBe(true);
+  });
+
+  it("saves under a name taken from the prompt, not the generation id", () => {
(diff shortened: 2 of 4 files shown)


sweetmantech:
## Preview verification — chat#1995

Verified on `chat-git-fix-music-download-attachment-recoup.vercel.app/music`, signed in as `sweetman@recoupable.com`, against 4 real generated songs in Supabase Storage.

### Root cause

The old link relied on the HTML `download` attribute. That attribute is **ignored for cross-origin URLs**, so Chrome fell back to a plain top-level navigation to the `.wav` — the tab left `/music` and rendered the browser's audio player. The fix moves the instruction server-side via Supabase Storage's `?download=<filename>` parameter.

### Documented vs actual

|---|---|---|---|
| 1 | Download href carries `?download=<slug>.wav` | All 4 cards, e.g. `...ca2e4b07….wav?download=genre-lo-fi-soul-bpm-82-key-d-minor-hazy-late-night-warm-tap.wav` | PASS |
| 2 | Supabase returns `Content-Disposition: attachment` | `attachment; filename=…; filename*=UTF-8''…` | PASS |
| 3 | Clicking download does **not** change the page route | URL before and after both `/music` | PASS |
| 4 | Page is not torn down and re-mounted | JS context survived the click (a sentinel set on `window` before the click was still present after) | PASS |
| 5 | Song list still rendered after download | 4 cards, 4 download links | PASS |
| 6 | Filename derived from the prompt, extension preserved | slug truncated to the 64-char budget, trailing hyphen trimmed, `.wav` kept | PASS |
| 7 | Unit suite green | 30/30 across `musicDownloadUrl`, `MusicGenerationCard`, `MusicGenerateForm`, `musicPricing` | PASS |

### The control that proves it

Same object, with and without the parameter:

```
$ curl -sI ".../ca2e4b07-….wav"
HTTP/2 200
content-type: audio/wav                       <- no disposition: browser navigates (the bug)

$ curl -sI ".../ca2e4b07-….wav?download=genre-lo-fi-soul-….wav"
HTTP/2 200
content-type: audio/wav
content-disposition: attachment; filename=genre-lo-fi-soul-bpm-82-key-d-minor-hazy-late-night-warm-tap.wav
```

The `attachment` disposition is what makes Chrome abort the navigation and hand the response to the download manager, which is why the tab stays put. `target="_blank"` was not needed — the fallback in the issue ("at least open a new tab") is unnecessary because the download never navigates in the first place.

### After clicking download — still on `/music`

Music page after clicking download

### Note on the test that let this ship

The previous assertion was `expect(download.hasAttribute("download")).toBe(true)`, which passed while the feature was broken — it checked that an attribute existed, not that it did anything. It is replaced with assertions on the resolved URL, which is the thing the browser actually acts on.

PR #1996, opened 2026-08-24
feat(music): song detail dialog with full prompt, lyrics, and settings

sweetmantech:
Closes the detail-dialog row of the chat#1992 PR matrix.

Clicking a song in **Your music** opens its full record.

## What it shows

| | |
|---|---|
| **Prompt** | In full. The card truncates it to one line, and seeing the whole of it is the reason this dialog exists. Copy button. |
| **Lyrics** | Every line, wrapped and scrollable within its own block so long lyrics do not push the settings off screen. Copy button. |
| **Settings** | Model, status, duration, seed. |
| **Failure reason** | Verbatim, for a failed generation. |

## Why four settings and not seven

The row asked for "full settings from the API response". Checking fal before building it, most of them do not survive the round trip: its result payload is `{ audio, seed, duration }`, its status endpoint returns only queue metadata, and `/input` and `/request` both 405.

`num_inference_steps` and `guidance_scale` are consumed at submit and never echoed anywhere, so they are not recoverable. Rather than mirror them into columns of our own, the dialog reports the settings that genuinely exist. `seed` comes from recoupable/api#850, which reads it live from fal.

## Interaction

The card is clickable, but the player and the download sit inside it, so their clicks stop at the controls row. Without that, pressing play or saving a song would also pop the dialog open. Both cases are covered by tests.

Keyboard users get a real `<button>` on the title rather than a `role="button"` on the card, which would have meant nesting the player and the download link inside a button.

The dialog opens populated from the summary the card already holds, so it does not show an empty shell while the seed is fetched. The detail read is enabled only while the dialog is open: it costs a fal round trip on the API side, and firing one per song on gallery load would be wasteful.

## Tests

18 new assertions, TDD (RED confirmed before implementing). 42/42 green across `components/MusicPage` and `lib/music`; `tsc --noEmit` and `eslint` clean.

Prompt and lyrics assertions compare against full textContent, so they fail if the dialog ever reuses the card's `truncate` treatment or slices the string.

Also fixes a genuine a11y gap surfaced while testing: Radix was warning about a `DialogContent` with no description, so the generation date now fills that role.

## Depends on

recoupable/api#850 for the `seed` field. Until it merges, `seed` reads **Not available** rather than breaking; the dialog treats a missing seed as a normal state.

---
## Summary by cubic
Adds a song detail dialog showing the full prompt, full lyrics, and run settings, and standardizes copy-to-clipboard with a shared `CopyButton`. Previously cards truncated text with no detail view; now clicking a song opens the dialog, and player/download clicks do not.

- Prompt and Lyrics render in full with per-block copy buttons; the dialog is the only scroll context (removed inner lyrics scroller).
- Settings show Model, Status, Duration, and Seed; omit `num_inference_steps` and `guidance_scale` because the provider never returns them.
- Seed loads on demand via `useMusicGeneration` (built on `@tanstack/react-query` and `@privy-io/react-auth`) calling GET /api/music/:id; shows Loading until settled, then a value or Not available; depends on `recoupable/api#850` but degrades gracefully.
- Accessibility/interaction: the title is a real button for keyboard access; the card opens the dialog on click; the player and Download stop propagation.
- Refactor: extract a shared `CopyButton` and use it in the dialog, `app/access`, and the API key modal; place `CopyButton` in its own folder with an index so imports remain `@/components/CopyButton`.

<sup>Written for commit e8c0d1fb34c97846334f38e2e83bbb35d6a21de6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1996?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added detailed music-generation views showing status, prompt, lyrics, settings, creation time, and failure details.
  * Added copy controls for prompts and lyrics with accessible labels and success feedback.
  * Music generation cards can now be opened via click or keyboard to view details.
  * Added display of duration and seed values, including unavailable-state messaging.
* **Bug Fixes**
  * Improved download behavior when audio is unavailable.


sweetmantech:
## Preview verification — chat#1996

Verified on `chat-git-feat-music-detail-dialog-recoup.vercel.app/music`, signed in as `sweetman@recoupable.com`, against 4 real generated songs. api#850 was merged into the api `test` branch first so the seed resolves for real rather than falling back.

### Documented vs actual

|---|---|---|---|
| 1 | Clicking a song opens the dialog | Opens on card click and on the title button | PASS |
| 2 | Prompt shown in full | 81/81 chars, byte-identical to the API payload | PASS |
| 3 | Lyrics shown in full | 424/424 chars, byte-identical to the API payload | PASS |
| 4 | Card title stays truncated | `text-overflow: ellipsis` on the card, full text in the dialog | PASS |
| 5 | Copy prompt | Copies all 81 chars, toast confirms | PASS |
| 6 | Copy lyrics | Copies all 424 chars, toast confirms | PASS |
| 7 | Settings from the API | model `minimax/music-3`, status `completed`, duration `0:56`, seed `1406002189` | PASS |
| 8 | Seed read live from fal | Real value end-to-end through api#850, not a stored column | PASS |
| 9 | Download click does not open the dialog | No dialog | PASS |
| 10 | Player click does not open the dialog | No dialog | PASS |
| 11 | Escape closes | Closes | PASS |
| 12 | Mobile 390px | Dialog 390px, no horizontal page overflow | PASS |
| 13 | Unit suite | 44/44 across `components/MusicPage` and `lib/music`; tsc + eslint clean | PASS |

Prompt and lyrics were compared against `GET /api/music` fetched from the page itself, so #2 and #3 are equality against the real payload rather than eyeballing.

### Two defects found and fixed during verification

**Lyrics were in a scroll box inside a scrolling dialog.** The block was capped at 192px while the lyric rendered 524px, so two thirds of it sat behind a second scrollbar that is easy to miss. That is the opposite of showing lyrics in full. The cap is gone and the dialog is now the only scroll context. Covered by a test that fails against the capped markup (verified by reverting the component alone, not the test).

**The seed claimed to be unavailable while it was still loading.** For the second or so before the detail read lands, the row read `Not available`, then flipped to a number. "Not available" is a verdict, so it now waits for the fetch to settle and reads `Loading` in the interim. Confirmed live: `Loading` immediately after open, `1406002189` after.

### Desktop — full prompt and full lyrics

Note the card behind the dialog, truncated to `Genre: lo-fi soul. BPM: 82. Key: ...`, against the same prompt in full.

Desktop dialog

### Desktop — settings

Desktop settings

### Mobile, 390x844

Mobile dialog

### Note on the api `test` branch

`test` currently carries api#850 ahead of its merge, so this preview could resolve a real seed. It normalizes once #850 lands on `main`.

PR #1997, opened 2026-08-24
fix(music): make the detail dialog closable

sweetmantech:
Reported on the chat#1992 matrix: the song detail dialog could not be dismissed. The X appeared to do nothing, and clicking away closed it and reopened it instantly.

## Cause

`MusicDetailDialog` was rendered as a **child** of the card `div` that carries `onClick={() => setDetailOpen(true)}`.

Radix portals the dialog to `document.body`, so in the DOM it is nowhere near the card — which is what makes this easy to miss. But **React events propagate along the React tree, not the DOM tree**. Every click inside the portal still reached the card's handler:

1. Click the X → Radix calls `onOpenChange(false)`
2. The same click bubbles up the React tree to the card → `setDetailOpen(true)`
3. Both land in one tick, so the dialog never visibly closes

Click-away was the same sequence, just with the reopen visible as a flash.

## Fix

The dialog is now a **sibling** of the card rather than a descendant, so the propagation path no longer exists. That is preferable to `stopPropagation` on the dialog content, which would have suppressed the symptom while leaving a card that swallows events from a subtree it should have no relationship with.

The bulk of the diff is the re-indent from nesting the card one level deeper inside a fragment; `git diff -w` shows the actual change is about fifteen lines.

## Tests

Three added, 47/47 green. TDD, with the two click paths confirmed RED first:

| Test | Before |
|---|---|
| closes when the X is clicked | **FAIL** |
| stays closed after a click inside the dialog | **FAIL** |
| closes on Escape | passed already |

Escape was never broken, because a keydown on `document` does not travel the React tree from the card. That is worth having as a test anyway, since it is the only close path that worked and a future refactor could take it away.

## Why the existing tests missed it

chat#1996 tested that clicking the download and the player do **not** open the dialog, but never that the dialog closes. I also verified the open path on preview and never tried to dismiss it. Both gaps are the same shape: the feature was checked for what it should not do on the way in, and not at all on the way out.

---
## Summary by cubic
Fixes the music detail dialog so it can be dismissed. Previously, clicking the X or outside immediately reopened the dialog; now X, click-away, and Escape close it and it stays closed.

**Review notes**
- Render MusicDetailDialog as a sibling of the card to break React event propagation that reopened it, avoiding stopPropagation on dialog content.
- Preserve player/download behavior by stopping propagation only within the card’s media row.
- Add tests: closes via X; stays closed after a click inside the dialog; closes on Escape.
- Most of the diff is indentation from wrapping the card in a fragment; the functional change is small.

<sup>Written for commit 53753432a0967abca5b147922ab9595d74aab31c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/1997?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/components/MusicPage/MusicGenerationCard.tsx
+++ b/components/MusicPage/MusicGenerationCard.tsx
@@ -20,94 +20,101 @@ const MusicGenerationCard = ({ generation }: { generation: MusicGeneration }) =>
   );
 
   return (
-    // min-w-0 because a grid item defaults to min-width:auto and will not
-    // shrink below its content. Without it this card rendered 719px wide in a
-    // 342px cell on mobile, pushing the player off screen.
-    <div
-      onClick={() => setDetailOpen(true)}
-      className="p-4 border rounded-lg min-w-0 cursor-pointer hover:bg-muted/40 transition-colors"
-    >
-      <div className="flex items-start justify-between gap-2">
-        <div className="min-w-0">
-          {/* A real button, so the dialog is reachable by keyboard. The card's
-              own onClick covers the rest of the surface for pointer users
-              without nesting interactive content inside a role="button". */}
-          <button
-            type="button"
-            aria-label={`View details for ${title}`}
-            className="block w-full text-left"
+    // The dialog is a sibling of the card, never a child of it. Radix portals
+    // it to document.body, but React events travel the React tree rather than
+    // the DOM tree, so as a child every click inside the dialog still reached
+    // this onClick and reopened it in the same tick Radix closed it, which
+    // made the X look inert and click-away reopen instantly.
+    <>
+      {/* min-w-0 because a grid item defaults to min-width:auto and will not
+          shrink below its content. Without it this card rendered 719px wide in
+          a 342px cell on mobile, pushing the player off screen. */}
+      <div
+        onClick={() => setDetailOpen(true)}
+        className="p-4 border rounded-lg min-w-0 cursor-pointer hover:bg-muted/40 transition-colors"
+      >
+        <div className="flex items-start justify-between gap-2">
+          <div className="min-w-0">
+            {/* A real button, so the dialog is reachable by keyboard. The card's
+                own onClick covers the rest of the surface for pointer users
+                without nesting interactive content inside a role="button". */}
+            <button
+              type="button"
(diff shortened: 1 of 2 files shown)


sweetmantech:
## Preview verification — chat#1997

Verified on `app-git-fix-music-dialog-close-recoup.vercel.app/music`, signed in as `sweetman@recoupable.com`, against the same four real songs the bug was reported on.

### The three close paths

| # | Path | Before | Actual | Result |
|---|---|---|---|---|
| 1 | Click the X | appeared inert | closes, stays closed, still on `/music` | PASS |
| 2 | Click away on the overlay | closed then reopened instantly | closes, stays closed | PASS |
| 3 | Escape | worked | still works | PASS |

Paths 1 and 2 were exercised with a **real trusted click** through the browser's input pipeline, not a scripted `.click()`, since the whole failure was about how a genuine click propagates. The X was clicked via the accessibility node; the overlay via a full `pointerdown → mousedown → pointerup → mouseup → click` sequence at (150, 450), landing on `DIV.fixed inset-0 z-50 bg-black/80` — the overlay itself, well clear of the dialog.

The reopen was the actual symptom, so each close is checked twice: immediately, and again 1.5s later. Both stayed closed.

### No regressions in what chat#1996 established

| Behaviour | Actual |
|---|---|
| Card click opens the dialog | opens |
| Download does not open it | does not open |
| Player does not open it | does not open |
| Seed still resolves from fal | `1406002189` |

That last group matters here: the fix moves the dialog out of the card, and the card's `stopPropagation` on the controls row had to keep working independently.

Song details dialog with a working Close button

### Why the original verification missed this

chat#1996's preview run opened the dialog and read its contents, and never tried to dismiss it. The unit tests had the same shape: they asserted what does **not** open the dialog and never that it closes. Both gaps pointed the same way, which is how a dialog that could not be closed passed a 13-point check.

PR #2001, opened 2026-08-24
refactor(copy): one copy component, CopyAction deleted

sweetmantech:
Row 7 of the chat#1999 matrix.

**Rewritten after review.** The first version moved `CopyAction`'s clipboard logic onto `useCopy` but kept the component. @sweetmantech's objection was correct: there is no reason to test and merge a refactor of something we were about to delete. This PR is the end state instead.

## Why CopyAction was not actually different

Its only structural difference was rendering `<Action>` rather than `<Button>`. Reading `components/actions.tsx`, **`Action` is `Button`** with `variant="ghost" size="sm"`, a `rounded-full` class, an `sr-only` label, and an optional Radix tooltip. A thin wrapper over the same primitive `CopyButton` already used.

Everything else that looked distinctive was a prop, not an architecture:

| | was in CopyAction | now |
|---|---|---|
| tooltip | hardcoded | `tooltip` prop |
| toast | suppressed | `silent` prop |
| icon size | `!w-3 !h-3` | `iconClassName` prop |
| container | `<Action>` | rendered when `tooltip` is set |
| motion icon rotation | framer-motion | **dropped** |

The motion is the one deliberate loss. It existed in exactly one control out of four copy surfaces, and the tick is the confirmation either way — keeping it would have meant carrying framer-motion into the shared component for a flourish nowhere else has.

## Result

```
grep -rn "clipboard" components hooks lib
  hooks/useCopy.ts                     ← the only implementation
  components/CopyButton/CopyButton.tsx ← the only control
  components/ArtistSetting/AccountIdDisplay.tsx  ← chat#2002
```

`AccountIdDisplay` is the last direct caller and is the next row, which branches from here.

## Tests

**496/496 green**, `tsc` and `eslint` clean. 5 new for `CopyButton`, including one pinning the toolbar variant — it asserts the rendered control keeps `rounded-full`, so a future change that quietly turns the chat toolbar's copy into a bordered button fails rather than shipping.

`useCopy` keeps the `silent` option and its first tests, which both this and chat#2002 need.

## Not preview-tested yet

The chat message toolbar is the busiest surface this touches and the one where a visual regression would be least welcome. Worth a look at an assistant message before merging — Copy should sit beside Retry and Edit with the same ghost, rounded, muted treatment, and show a tick with no toast.

## Summary by CodeRabbit

- **Enhancements**
  - Standardized copy controls across message actions with consistent icons, tooltips, and copied-state feedback.
  - Added an option to copy content without showing a success notification.
  - Clipboard errors continue to display error notifications.
  - Copy controls now support customizable icon styling and toolbar action presentation when tooltips are provided.


sweetmantech:
## Preview verification — chat#2001

Verified on `app-git-refactor-copy-action-dry-recoup.vercel.app`, on a real chat with a real assistant reply. This is the surface `CopyAction` used to own, so the whole point was to look at it rather than trust the unit tests.

### It sits level with its neighbours

The concern with moving from `Action` to a `Button` plus call-site classes was that Copy would drift from Retry and Edit. Measured on the deployed build, in the same toolbar:

| control | size | radius | colour | icon |
|---|---|---|---|---|
| Retry / Edit | 32x32 | pill | `rgb(161,161,170)` | 12x12 |
| **Copy** | **32x32** | **pill** | **`rgb(161,161,170)`** | **12x12** |

Identical on every axis, including the `!w-3 !h-3` icon override landing over the `[&_svg]:size-4` base.

### Documented vs actual

|---|---|---|---|
| 1 | Copy renders on assistant messages | present, beside Retry | PASS |
| 2 | …and on user messages | present, beside Edit | PASS |
| 3 | Matches neighbour chrome | identical size, radius, colour, icon | PASS |
| 4 | Copies the assistant's response | clipboard received `"ok"` | PASS |
| 5 | Copies the user's message from its own toolbar | received `"Reply with exactly: ok"` | PASS |
| 6 | Tick replaces the copy icon | icon swapped | PASS |
| 7 | …and resets after the delay | back to the copy icon | PASS |
| 8 | **No toast** | none shown | PASS |
| 9 | Tooltip on hover | "Copy response to clipboard" | PASS |
| 10 | Unit suite | 497/497, tsc + eslint clean | PASS |

**#8 is the one that would have been a silent UX change.** `useCopy` toasts by default; this control never did. Confirmed absent on the deployed build, not just in a mock.

**#9 nearly became a false bug report.** My first hover check found no tooltip and I was ready to call it a regression from dropping `Action`. Before doing that I hovered Retry as a control — its tooltip appeared, proving the harness could trigger Radix — then re-tested Copy and the tooltip was there. The first sample was simply too early. Worth recording because the wrong conclusion was one step away, and the control experiment is what prevented it.

### Screenshot

Copy beside Retry under an assistant reply, tooltip open, hover state on:

Chat toolbar

### The tradeoff this locks in

Copy's visual match to Retry and Edit is now maintained **by hand** at the `MessageParts` call site rather than inherited from `Action`. Today they are pixel-identical, as measured above. If that toolbar is ever restyled, this call site has to move with it — there is a comment pointing at `components/actions.tsx`, but a comment is weaker than construction. That is the accepted cost of one render path instead of a branch.

Related: the test that used to assert `rounded-full` came from the component now asserts className passthrough, since the styling moved to the caller. The "matches its neighbours" guarantee is a call-site convention, not test-enforced.

PR #2003, opened 2026-08-24
feat(music): quote the price as an upper bound at fal's rate

sweetmantech:
Row 8 of the chat#1999 matrix. The client half of recoupable/api#853.

## Two changes

**Rate.** `CREDITS_PER_SECOND` 0.5 → 0.2 and the 15-credit floor removed, matching the API exactly. fal charges $0.002/s and a credit is $0.01, so 0.2/s is fal's rate with no markup.

| duration | before | after |
|---|---|---|
| 10s | $0.15 | **$0.02** |
| 60s | $0.30 | **$0.12** |
| 120s | $0.60 | **$0.24** |
| 300s | $1.50 | **$0.60** |

**Wording.** `Costs $0.12 for 60s (12 credits)` → `Up to $0.12 for 60s (12 credits). You are charged for the audio actually generated.`

api#853 charges for the audio fal actually produced, and the model routinely stops short — our first songs averaged 39.9s against a 60s default. A fixed "Costs" would overstate what most generations are billed. An upper bound is safe in the only direction that matters: nobody is charged more than they were shown.

This is why the earlier "the quoted price must be final" reasoning no longer applies. It existed to stop the quote being a lie; the charge now only ever moves in the customer's favour.

## On the duplicated constant

`lib/music/const.ts` restates the API's rate rather than fetching it, because the quote has to re-render on every slider drag. That was a deliberate call when the pricing shipped and it still holds — but it means **this PR and api#853 have to land together or the form lies**. The tests now name api#853 so the next person changing one finds the other.

## Tests

47/47 green, both suites updated first and confirmed RED. `tsc` and `eslint` clean.

The form test now asserts the "up to" phrasing and the explanatory line, not just the number, so dropping the caveat while keeping the price would fail.

---
## Summary by cubic
Quotes music generation as an upper bound at fal’s rate and aligns the UI with the API. Previously: “Costs $X” at 0.5 credits/s with a 15-credit floor; now: “Up to $X” at 0.2 credits/s with no floor, and we state billing is for audio actually generated (e.g., 60s now shows $0.12).

- Rollout
  - Deploy together with the API change (recoupable/api#853); otherwise the form quote will not match charges.
  - The rate is duplicated in `lib/music/const.ts` for responsive quoting; keep it in sync with the API when pricing changes.
  - Tests now assert the “Up to …” wording and the explanatory line.

<sup>Written for commit 02f4e81be618f01d9826db3f50ab29aae7e7d1d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/chat/pull/2003?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/components/MusicPage/MusicGenerateForm.tsx
+++ b/components/MusicPage/MusicGenerateForm.tsx
@@ -85,7 +85,8 @@ const MusicGenerateForm = () => {
 
         <div className="flex flex-col-reverse gap-2 sm:flex-row sm:items-center sm:justify-between">
           <span className="text-xs text-muted-foreground">
-            Costs {formatCreditCostUsd(credits)} for {settings.duration}s ({credits} credits)
+            Up to {formatCreditCostUsd(credits)} for {settings.duration}s ({credits} credits).
+            You are charged for the audio actually generated.
           </span>
           <div className="flex gap-2">
             <Button variant="ghost" onClick={reset} disabled={isPending} className="flex-1 sm:flex-none">
--- a/components/MusicPage/__tests__/MusicGenerateForm.test.tsx
+++ b/components/MusicPage/__tests__/MusicGenerateForm.test.tsx
@@ -43,10 +43,14 @@ describe("MusicGenerateForm", () => {
     );
   });
 
-  it("quotes the price in dollars with credits secondary", () => {
+  it("quotes the price as an upper bound, in dollars with credits secondary", () => {
+    // "Up to", because the API charges for the audio fal actually produced and
+    // the model routinely stops short (recoupable/api#853). A fixed "Costs"
+    // would overstate what most generations are billed.
     setup();
 
-    expect(screen.getByText(/Costs \$0\.30 for 60s \(30 credits\)/)).toBeDefined();
+    expect(screen.getByText(/Up to \$0\.12 for 60s \(12 credits\)/)).toBeDefined();
+    expect(screen.getByText(/charged for the audio actually generated/i)).toBeDefined();
   });
 
   it("sends the documented defaults and omits a blank seed", () => {
@@ -83,7 +87,7 @@ describe("MusicGenerateForm", () => {
 
     fireEvent.change(screen.getByLabelText("Duration"), { target: { value: "120" } });
 
-    expect(screen.getByText(/Costs \$0\.60 for 120s \(60 credits\)/)).toBeDefined();
+    expect(screen.getByText(/Up to \$0\.24 for 120s \(24 credits\)/)).toBeDefined();
   });
 
   it("fills the seed from the randomize button", () => {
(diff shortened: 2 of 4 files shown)


sweetmantech:
## Preview verification — chat#2003

Verified on `app-git-feat-music-price-upper-bound-recoup.vercel.app/music`, against the **already-shipped** api#853.

### The drift this closes, measured on production

api#853 merged, so production is now charging the new rate while the form still quotes the old one. Read off `chat.recoupable.dev/music` just now:

```
Costs $0.30 for 60s (30 credits)
```

The API charges **at most 12 credits** for that request, and less when the model stops short — one of yesterday's test songs asked for 60s, produced 15.6s, and cost **4**. So production currently overstates by 2.5x at best and 7.5x at worst. This PR is the fix and should follow api#853 promptly.

### Documented vs actual

|---|---|---|---|
| 1 | Quote reads as an upper bound | `Up to $0.12 for 60s (12 credits).` | PASS |
| 2 | Caveat explains why | `You are charged for the audio actually generated.` | PASS |
| 3 | Re-prices as the slider moves | 6 durations checked live, all correct | PASS |
| 4 | Matches the shipped API exactly | agrees at every point tested | PASS |
| 5 | No floor | 10s quotes 2 credits, not 15 | PASS |
| 6 | Renders before sign-in | correct while signed out | PASS |
| 7 | Mobile 390px | fits, no horizontal overflow | PASS |
| 8 | Unit suite | 47/47, tsc + eslint clean | PASS |

**#4 is the one that matters**, since the rate is deliberately duplicated client-side so the quote can re-render on a slider drag. Every value the form shows was compared against `creditCostForDuration` as it exists on api `main` after api#853:

| duration | form quote | API charge | agree |
|---|---|---|---|
| 10s | 2 | 2 | ✓ |
| 20s | 4 | 4 | ✓ |
| 30s | 6 | 6 | ✓ |
| 60s | 12 | 12 | ✓ |
| 120s | 24 | 24 | ✓ |
| 300s | 60 | 60 | ✓ |

Exact agreement across the full 10–300s range, including both ends.

### Screenshots

**Desktop, Additional Settings open at the 60s default:**

Desktop quote

**Mobile, 390x844:**

Mobile quote

### Note

The quote is an upper bound by design: the API charges the audio fal actually produces, so the real deduction is usually lower. Verified against a live generation on api#853 — 60s requested, 15.6s produced, 4 credits charged against the 12 quoted. The promise only ever moves in the customer's favour.

Issue #2052, opened 2026-09-02
Music videos in chat: the render chain as a skill, and media components for the sandbox agent


sweetmantech:
**Shipped and closed 2026-09-04.** Recoup Chat can now produce a music video in one conversation: the agent generates the song, drives the stills and the motion through pinned house models, and the results play inside the thread. When this issue opened, every building block existed in `api` but the agent had no encoded recipe for chaining them and the app could not render what came back — both are now closed.

Eleven PRs across four repos, listed below. Two follow-ups were carried out rather than shipped here, and live in [app#2059](https://github.com/recoupable/app/issues/2059): a generated **song** still does not render inline (`POST /api/music` answers `202` with `audio_url: null`, and the skill's polling loop never prints the completed body to stdout), and the sandbox base snapshot lacks Chrome's runtime libraries, so the render step installs them mid-run. The **free-tier duration cap** in the original title was never built and moved there too, along with the remaining tool components and the chat-UI cleanup.

> **Correction, 2026-09-02 (Sweets).** The first draft of this issue opened with a `generate_music` MCP tool. That was the wrong layer. Chat runs the **sandbox agent** with a deliberately minimal tool set and reaches Recoup over **HTTP**, so an MCP tool would have shipped music to every surface except the chat. Decided: no new MCP tools; the chain is taught as a skill that curls the existing endpoints. See Architecture decisions.

## Goal

A user opens a chat, says "make a music video for this song" (or "…for a track about X"), and gets a finished music video back in the thread — playable inline, downloadable, rendered over real full-length audio.

What that turned into, and where each part landed:

- **Song** ✅ — the agent curls the existing `POST /api/music` (MiniMax Music 3 via fal, workflow-polled, pass-through priced).
- **Visuals** ✅, on a narrower path than first drafted — `POST /api/content/image` (Muse Image) and `POST /api/content/video` (MiniMax H3 Max), both pinned server-side and credit-gated. The `transcribe` / `PATCH /api/content` edit-op route was superseded: those endpoints had been cold since April and `transcribe` 500s on every input, so the skill builds on the hyperframes + ffmpeg chain instead (see Architecture decisions).
- **Surface** ✅ for video and stills — they play on the tool row that produced them ([app#2057](https://github.com/recoupable/app/pull/2057)). ⚠️ **Not** for the song: see the lead.
[...]

PR #2057, opened 2026-09-02
feat(chat): play generated media on the tool row that produced it

sweetmantech:
Row 5 of #2052 — the payoff row.

> **Stacked on #2056 → #2055.** Retarget to `main` as those merge.

## Why

A finished song or video arrives in the thread as a **blue hyperlink**. The most important moment in the product requires leaving the conversation to find out whether the thing you asked for is any good.

## Surface decision

Reads the **bash tool result**, not the assistant's prose. An earlier draft of this row parsed media URLs out of markdown; that depends on the agent's prose discipline, needs an origin allowlist to avoid embedding third-party URLs it scraped via `web_fetch`, and renders media wherever the agent happens to mention it rather than where it was produced.

**Keys off stdout, not the command params.** `POST /api/music` returns `202` with a generation id — the audio URL only appears later in the `GET /api/music/{id}` poll response. "The command contained `/api/music`" tells you a generation *started*, not where the media is.

## Security

Only well-known response fields from **our own API** are read (`audio_url`, `videoUrl`, `videoSourceUrl`, `imageUrl`, …), and the value must be an absolute http(s) URL with a media extension. A URL the agent echoed from a scraped page — even in the same JSON blob — is never rendered as a player. Two tests pin exactly that.

## Behaviour

The player is added **beneath** the row, never replacing it: a failed call still needs its row, and the command that produced a file is worth seeing. Nothing renders for a failed call.

Video is height-capped rather than width-capped, since a music video is 9:16 and a width-constrained portrait clip would fill the whole thread.

## Depends on the skill row

The agent may pipe through `jq`, redirect to a file, or poll in a loop, in which case the URL never reaches stdout. The music-video skill (row 10) must guarantee its final step prints the media URL to stdout. Designed together.

## Tests

8 cases on `extractMediaFromStdout`: the music poll shape, nested run output, images, query-string classification, **two security cases** (unknown field, non-media extension), non-JSON stdout, and relative URLs. Full run: **55 passed**. `tsc --noEmit` and eslint clean.

## Preview verification

Pending — draft until verified on the preview for this SHA. Done when: a chat that generates a song renders an inline audio player under the bash row that produced it, playable and downloadable; a generated video renders with correct aspect; a failed generation still shows its bash row with the error. Dark, light and mobile screenshots.

https://claude.ai/code/session_01KcNQAfA3nS7BTrMB3bXsnX

---
## Summary by cubic
Implements row 5 of #2052: generated audio, video, and images now play inline under the bash tool row that produced them, instead of arriving as a blue hyperlink in the thread. Also replaces the chat's three separate download implementations with one that actually saves the file.

**Media extraction**
- Reads stdout JSON but only known response fields, so URLs the agent echoed from scraped pages never render as players.
- The player sits beneath the row; failed calls still show their row and no media.
- Video is height-capped, since a 9:16 clip would otherwise fill the thread.
- The music-video skill must print the media URL to stdout; piped or redirected output won't render.

**Downloads**
- Downloads fetch the bytes and save through an object URL, because the `download` attribute is ignored cross-origin; falls back to opening the URL directly when the fetch fails.
- One `useMediaDownloader` replaces three hand-rolled paths; prefetch is now opt-in and only the image tool uses it.
- The dead Sora2 video surface is deleted — the chat agent can't emit that tool, and its download button never saved the file.
- The image tool keeps its date-stamped filename.
- Adds 8 extraction tests plus download tests for the object-URL path and fallback.

<sup>Written for commit bab9af026b05ff8096b8573359b1e49f5721119d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/recoupable/app/pull/2057?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Bash tool results can now display extracted audio, video, and image content inline.
  - Media remains visible whether the tool output is expanded or collapsed.
  - Added playback and preview controls, along with an option to download displayed media.
  - Downloads now support audio, video, and images with improved filename handling and download states.

- **Changes**
  - Removed dedicated Sora 2 video retrieval result displays.


components/VercelChat/tools/agent/MediaResult.tsx
sweetmantech, line 1: SRP
- actual: all media types are inlined in the MediaResult component
- required: child components for each media type

